PR #21, opened 2021-04-02
Ois curve

alecloudenback:


alecloudenback:
Tests failing - I think it's because the coupons after the last fixed point are getting the wrong rate implied by interpolation. Need to recalculate interpolation iteratively to converge on intended rate for interim coupons. Issue may also affect `USTreasury` but not surfaced in tests due to limted number of digits avaialable to compare.

alecloudenback:
closed in lieu of #25

Issue #23, opened 2021-04-04
provide function for efficient vector of accumulation/discount


alecloudenback:
e.g. `discounts` and `accumulations` would return a vector filled with discount/accumulation factors rather than needing to call and accumulate rates with multiple `discount` calls.

alecloudenback:
I think I ran into some sort of disambiguation error when doing `discount(curve,[1,2,3])` but it seems to work (mostly, see #40). Should also be able to optimize based on the second argment being an `AbstractVector`.

Todo: 
- [ ] check tests and see if this is tested and what the behavior is on the different curve types.

PR #37, opened 2021-10-02
Smith Wilson yield

kasperrisager:
Example of the Smith-Wilson yield curve included into Yields.jl. This is a draft for discussing ideas, so it's a little light on testing, doc and performance optimizations.

It thrown together from code I had lying around, and I didn't make a big effort to make it look Yields.jl'ish. The `CalibrationInstruments` struct makes a lot of sense for the SW curve, since the calibration neatly reduces to a linear algebra problem, but for e.g. bootstrapping, the structure would have to be different (as a start, you would want instruments ordered by expiry).


alecloudenback:
I don't have substantive feedback to give yet - trying to first follow what SW is doing and found the following resources which could be sources of additional information and test cases:

- [Smith-Wilson Yields Curves](http://gli.lu/2017/12/smith-wilson-yield-curves/)
- [A Technical Note on the Smith-Wilson Method](http://www.ressources-actuarielles.net/EXT/ISFA/fp-isfa.nsf/2b0481298458b3d1c1256f8a0024c478/bd689cce9bb2aeb5c1257998001ede2b/$FILE/A_Technical_Note_on_the_Smith-Wilson_Method_100701.pdf)

alecloudenback:
I translated the linked example above into Julia to help me follow along. I still need to digest what each step was doing, as thus far it was more of a translation exercise.

src/smith_wilson.jl
```diff
@@ -0,0 +1,135 @@
...
+    discount(swy::SmithWilsonYield,t)
+
+Discount factor for a Smith-Wilson yield curve.
+"""
+function discount(swy::SmithWilsonYield, t)
+    return exp(-swy.ufr * t) * (1.0 + sum([H(swy.alpha, swy.u[midx], t) * swy.qb[midx] for midx in 1:length(swy.u)]))
+end
+
+# Utility method - this could be extended to all YieldCurves
+discount(swy::SmithWilsonYield, av::AbstractVector) = [discount(swy, t) for t in av]
```
alecloudenback, line 47: I definitely like the idea of providing this result. In #23 I had thought that the function would be called `discounts` to disambiguate when a vector would be returned vs just a scalar but it seems to mostly work (#40) already.

src/smith_wilson.jl
```diff
@@ -0,0 +1,135 @@
...
+"""
+    CalibrationInstruments(t, cf, p)
+
+Cash flows for calibrating a yield curve, along with their prices and payment times.
+"""
+struct CalibrationInstruments
+    t    # Column vector of maturities
+    cf   # Matrix of cash flow for each [maturity, instrument]
+    p    # Row vector of instrument prices
+end
```
alecloudenback, line 58: This doesn't seem to be used aside from constructing a SW curve, nor is it used for dispatch. Would it be simpler or easier to remember to pass a tuple or named arguments instead?

alecloudenback:
> but for e.g. bootstrapping, the structure would have to be different (as a start, you would want instruments ordered by expiry).

My understanding is that one of the key advantages of SW is that it's analytically solved with no boostrapping. When would this be combined with bootstrapping?


Some things to do before merging:

- [ ] Add some more test cases (the links I posted above have some numeric test cases)
- [ ] Add `accumulation` function (can just be inverse of `discount`)
- [ ] Add note on the ReadMe about this functionality

src/smith_wilson.jl
```diff
@@ -0,0 +1,135 @@
+using LinearAlgebra
+
+export SmithWilsonYield, CalibrationInstruments
+export InstrumentQuotes, ZeroCouponQuotes, SwapQuotes, BulletBondQuotes
+
+"""
+    SmithWilsonYield(ufr,alpha,u,qb)
```
alecloudenback, line 7: I think it might be better to just call this `SmithWilson`. I don't think there's an ambiguity issue, and reads better when imported elsewhere, e.g.:

```julia
import Yields

Yields.SmithWilson(...)
```

alecloudenback: Also, the literature seems to be so consistent in the use of `α` and Julia make it so easy to enter unicode, I think we should just use `α` instead of `alpha`

alecloudenback:
This is sparking a lot of really great ideas, thanks for the PR again!

src/smith_wilson.jl
kasperrisager: Agreed, I missed that the broadcast behaviour already did this.

src/smith_wilson.jl
kasperrisager: My original intent was that there would be an `AbstractCalibrationInstruments` where you could also do things like iterate over the cash flows and put that into a bootstrapping algorithm. But I agree that until that exists, it's going to look over-engineered to have a struct for just the sw curve.

alecloudenback: I think it brings up the question about pulling basic asset instruments into its own set of types and/or package? There's a lot more that could be explored there, like basic asset modeling. I don't think we have to hold up SW yields for that, but it could be a near term feature to build out!

alecloudenback:
A few tweaks:

- prefer full name `frequency` to `freq` for consistency with rest of package 
- use dispatch to make calls to `cashflows` less verbose by dispatching on the quoted instruments
- split up `bullet_cashflows` (which reutrns a tuple of `(times,cashflows)` into two functions `cashflows` and `timepoints`. This is to be consistent with the API in [LifeContingencies.jl](https://github.com/JuliaActuary/LifeContingencies.jl), which uses `cashflows` and `timepoints` to effectively do the same thing on some liability structs.

One solid request and then a question before merging:

1. Request: If you could just add some more detail to the docstring for `SmithWilsonCurve` that would help orient users as to how to construct with the pieces.
2. Question: `ufr` and `α` aren't used for dispatch, and the I found hard to remember the order of argumnets, which makes the two good candiates for being keyword arguments. So the API when using the quoted instruments would be: `SmithWilson(swq, α=0.1,ufr=0.05)` instead of `SmithWilson(0.1,0.05,swq)`. What do you think about making that change?

alecloudenback:
I made a few final tweaks to docstrings and made the `u`/`qb` method use keyword args as well for API consistency. Locally I couldn't get the generated docs to show the new methods. I'm hoping that the Github actions will catch them and build it into the Dev Docs.

PR #56, opened 2021-11-11
Let Rate do the trivial interest manipulations

kasperrisager:
The trivial interest manipulations are moved from depending on Constant to depending on Rate

### DIFF
--- a/src/Yields.jl
+++ b/src/Yields.jl
@@ -166,6 +166,31 @@ function rate(r::Rate{<:Real,<:CompoundingFrequency})
 end
 
 """
+    accumulation(r::Rate, t)
+    accumulation(r::Real, t)
+    accumulation(r::Union{Rate, Real}, from, to)
+
+Returns the accumulation factor for the rate `r` over the time `t`, or over the interval between `from` and `to`.
+"""
+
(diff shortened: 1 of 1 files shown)


kasperrisager:
Purpose is to move the responsibility for trivial interest accumulation to depend on `Rate`. Before this PR it went through `Constant<:AbstractYield`, but this makes it difficult to untangle simple rates from yield curves. With this inverted dependency, rate logic can be separated from yield curve logic.

alecloudenback:
Is this still a draft/WIP?

kasperrisager:
Wanted to know if there was appetite for going in this direction before finishing up a proper pr.

alecloudenback:
I'm not opposed to it - it seems like there's a degree of API consistency of technically `discount` is only defined on `AbstractYield`s and`Rate`s are used to construct yield curves. 

However, I do see how folks would want to use discounting on rates (thus the promotion to `Constant` yield curve. 

From a performance standpoint, Julia might inline the promotion and so there may or may not be a performance benefit from skipping `Constant`

kasperrisager:
Superseded by #80.

PR #65, opened 2021-11-20
add `ForwardStarting` to allow creating forward-starting curves easily

### LINKED ISSUE #63: New function: `forwardcurve`
`forwardcurve(YieldCurve,time)` would construct a new `YieldCuve` implied by the old one starting at `time`.

@kasperrisager any thoughts on the best way to accomplish this with `SmithWilson`? I think the others I can picture how this would be written but my inexperienced SW view is that it would be more difficult to do?

alecloudenback:
Closes #63

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,23 @@ y1 = Yields.Constant(0.05)
 y2 = y1 + 0.01                # y2 is a yield of 0.06
 ```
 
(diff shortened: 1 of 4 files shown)


kasperrisager:
Had a review pending already, but I'd forgotten to submit, apparently 😀

PR #66, opened 2021-11-20
Improve the display of curves

alecloudenback:
With the new parameterized types, the display of `AbstractYields` has gotten very verbose.
Unicode plots is pretty lightweight dependency.

Before:

```julia-repl
julia> zs = [5.0, 5.8, 6.4, 6.8] ./ 100
julia> maturity = [0.5, 1.0, 1.5, 2.0]

julia> curve = Yields.Zero(zs, maturity)
Yields.YieldCurve{Vector{Float64}, Vector{Float64}, Interpolations.Extrapolation{Float64, 1,
Interpolations.GriddedInterpolation{Float64, 1, Float64, Interpolations.Gridded{Interpolations.
Linear{Interpolations.Throw{Interpolations.OnGrid}}}, Tuple{Vector{Float64}}}, Interpolations.
Gridded{Interpolations.Linear{Interpolations.Throw{Interpolations.OnGrid}}}, Interpolations.
Line{Nothing}}}([0.05, 0.057999999999999996, 0.064, 0.068], [0.5, 1.0, 1.5, 2.0], 5-element
extrapolate(interpolate((::Vector{Float64},), ::Vector{Float64}, Gridded(Linear())), Line())
with element type Float64:
 1.0
 0.9759000729485331
 0.9451795841209829
 0.9111451296164018
 0.8767130973923045)
 ```

 After:

image

To consider further:

- [x] How to determine the ending point to display? 

- E.g. the above errors for `30` as endpoint because the discount factor hits zero so far before then and the `zero` hits a negative number and results in a complex result.
- Some curves would have a natural end point of the last defined observation time, but we'd still have to define a fallback time (e.g. a `Constant` curve).

> 30 seems as good as any timepoint for now, can add more specificity or get more dynamic about it as future enhancements

- [x] Show the ylim as (0,max(curve)) but without computing all the points we may not know the limits (plus could go negative)

- [x] Is there a better summary statistic to show than zero rates? Most people won't be familiar with looking at discount or accumulation curves.

> Zero rates seem fine for consistency across different kinds of curves for now. In the future, could try to inpsect the type of rates that were originally used to construct the curve and display them?

- [ ] Is there a way to test?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -7,6 +7,7 @@ version = "0.8.0"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 Interpolations = "a98d9a8b-a2ab-59e6-89dd-64a1c18fca59"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+UnicodePlots = "b8865327-cd53-5732-bb35-84acbb429228"
 
 [compat]
 ForwardDiff = "^0.10"
--- a/README.md
+++ b/README.md
@@ -95,6 +95,23 @@ y1 = Yields.Constant(0.05)
 y2 = y1 + 0.01                # y2 is a yield of 0.06
 ```
 
+### Forward Starting Curves
+
+Constructed curves can be shifted so that a future timepoint becomes the effective time-zero for a said curve.
+
+```julia-repl
+julia> zero = [5.0, 5.8, 6.4, 6.8] ./ 100
+julia> maturity = [0.5, 1.0, 1.5, 2.0]
+julia> curve = Yields.Zero(zero, maturity)
+julia> fwd = Yields.ForwardStarting(curve, 1.0)
+
+julia> discount(curve,1,2)
+0.9275624570410582
+
+julia> discount(fwd,1) # `curve` has effectively been reindexed to `1.0`
+0.9275624570410582
+```
+
 ### Conversion
 
 Convert rates between different types with `convert`. E.g.:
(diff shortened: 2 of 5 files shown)


kasperrisager:
This would be a really cool feature to have! Would help a lot in a REPL workflow.

From a pure software engineering perspective, I would suggest that the code is encapsulated in a plot-like method with more degrees of freedom (but reasonable defaults for everything) and then have Base.show point to that method with all default choices. That way, part of the layout choices can be used for other plots and not just for Base.show.

kasperrisager:
+ Splitting the code into one part that does layout choices and one part that forwards to a plotting library will make testing easier.

alecloudenback:
Updated look:

image

> From a pure software engineering perspective, I would suggest that the code is encapsulated in a plot-like method with more degrees of freedom (but reasonable defaults for everything) and then have Base.show point to that method with all default choices. That way, part of the layout choices can be used for other plots and not just for Base.show.

I think this is basically descriibing a [recipe](https://github.com/JuliaPlots/RecipesBase.jl), however it doesn't seem like UnicodePlots supports recipies by itself (it should if used as a Plots.jl backend). However, Plots.jl is too heavy of a dependency for this purpose. I propose merging as-is and tweaking in the future. I think this is effectively an improvement from the status quo and can be tweaked over time. @kasperrisager is that cool?

alecloudenback:
It's actually not working outside of REPL like I thought it would (e.g. a simple `lineplot` in Pluto will dispaly a plot in the notebook, but this PR does not). I [asked for help](https://julialang.zulipchat.com/#narrow/stream/243342-pluto.2Ejl/topic/Custom.20show.20methods.20for.20notebooks/near/262662458) with it on Zulip.

kasperrisager:
> @kasperrisager is that cool?

Sure, no reason to hold that feature back 🙂

PR #97, opened 2022-04-04
An initial attempt to add NelsonSiegelSvensson and SmoothingSpline but may be discussed further

leeyuntien:
Approaches taken to estimate parameters may be discussed further.


kasperrisager:
Thanks for taking a shot at the interpolations! It's a bit mouthful to start with, so we'll probably have a good and long discussion about it 😄. I guess Alec will join as reviewer, but until then, I can give some immediate feedback to keep you going. I'll mostly focus on the Nelson-Siegel part, but my comments may apply to some of the others, also.

1. For all of the curves, you would want to implement the method for the zero-coupon rate, `zero` and the discount function `discount`. I will also make the fitting procedures more readable if you refer directly to those, e.g., instead of defining it in the `spot` function inside the Nelson-Siegel fitting.
2. The `r^2` field on `NelsonSiegel` is not needed for `zero` or `discount`, so no need to keep it there. Not that it could not be interesting, but it's more a property of the fitting than of the result.
3. Looks like you're fitting the NS spot rate to the `prices` variable in input. That represents a bond price or a discount factor, so it should be converted back to an interest rate.
4. Just by googling, I can see that NS fitting is a big research topic, and there are probably many ways to go. Your method of using a fixed grid would probably work fine, but maybe it would be better to make the grid (and starting point?) inputs to the fitting function? That way, the user has more freedom.
5. We need to have tests added, and have the code integrated properly into the rest of the package, but let's take that down the line. You can use SmithWilson as an example of how to do this.

leeyuntien:
Thank you @kasperrisager for the comments! I agree with you on points 1, 2 and 4 and I have added zero and discount functions and made changes to the structs and functions by removing r square and by moving the fixed grid to the parameter list, but I did not fully get what point 3 means. I have checked the implementation of SmithWilson and found some variants also refer to prices, so it would be great if you can elaborate more on this?

kasperrisager:
Sure! What you have in line 65 is the formula for the (spot) interest rate for each maturity (given params and tau). However, when you do the fit in line 67, you adjust `params` so that the vector of interest rates is close to the vector `prices`. If `prices` comes from `ZeroCouponQuote`, they represent discount factors and not interest rates, so you're comparing two different things - either you have to transform the interest rates into discount factors or transform prices into interest rates before doing the fit.
I'm not an expert on Nelson-Siegel fitting, but it feels best to me to fit it to interest rates, in which case `prices` needs to be transformed. For the Smith-Wilson (where I do consider myself an expert 😄) the fit is done on discount factors, so that can naturally cause some confusion.

kasperrisager:
Looking forward to the next version, btw!

leeyuntien:
I have committed a change to a source file in this stream of pull request, but if any other steps i should also follow please advise. Thank you and look forward to discussing with you further.

src/NelsonSiegelSvensson.jl
```diff
@@ -0,0 +1,179 @@
...
+    #function NelsonSiegelSvensson(β₀, β₁, β₂, β₃, τ₁, τ₂, r²)
+    function NelsonSiegelSvensson(β₀, β₁, β₂, β₃, τ₁, τ₂)
+        #(τ₁ <= 0 || τ₂ <= 0 || r² < 0 || r² > 1) && throw(DomainError("Wrong parameter ranges"))
+        (τ₁ <= 0 || τ₂ <= 0) && throw(DomainError("Wrong parameter ranges"))
+        #return new(β₀, β₁, β₂, β₃, τ₁, τ₂, r²)
+        return new(β₀, β₁, β₂, β₃, τ₁, τ₂)
+    end
+end
+
+Base.zero(ns::NelsonSiegel, t) = ns.β₀ .+ ns.β₁ .* (1.0 .- exp.(-t ./ ns.τ₁)) ./ (t ./ ns.τ₁) .+ ns.β₂ .* ((1.0 .- exp.(-t ./ ns.τ₁)) ./ (t ./ ns.τ₁) .- exp.(-t ./ ns.τ₁))
```
kasperrisager, line 61: `Base.zero` should return a `Rate`, and for Nelson-Siegel and friends, that should be a continuous compounding rate. More concretely, the right-hand-side should be wrapped in a `Continuous(   )`.

kasperrisager: But then in the definition of `discount`, `zero` has to be converted back into a real number (rather than a `Rate`).

src/NelsonSiegelSvensson.jl
```diff
@@ -0,0 +1,179 @@
...
+discount(nss::NelsonSiegel, t) = exp.(-t .* zero(nss, t))
+
+using LsqFit
+
+""" 
+    est_ns_params(swq::Vector{SwapQuote}, τₐₗₗ::Array{Float, 1})
+
+Return the NelsonSiegel fitted parameters.
+"""
+function est_ns_params(swq::Vector{SwapQuote}, τₐₗₗ::Array{Float64, 1} = [0.1, 0.15, 0.2, 0.3, 0.5, 0.75, 1, 1.5, 2, 3, 5, 7.5, 10]) where {Q<:ObservableQuote}
```
kasperrisager, line 74: This function ends up very similar to the one below. As a general rule, there should not be any code duplication. I think the best solution right now is to just keep the version below that takes yields and maturities as `AbstractArray`s. Then later, it would be easy to add a method that extracts interest rates from quotes. Also, using `SwapQuotes` in this case is a little dangerous, since swaps are usually quoted as par rates and not spot rates.

src/NelsonSiegelSvensson.jl
```diff
@@ -0,0 +1,179 @@
...
+
+    return ns_param
+end
+
+""" 
+    est_ns_params(yields::AbstractVector, maturities::AbstractVector, τₐₗₗ::Array{Float64, 1})
+
+Return the NelsonSiegel fitted parameters.
+"""
+function est_ns_params(yields::AbstractVector, maturities::AbstractVector, τₐₗₗ::Array{Float64, 1} = [0.1, 0.15, 0.2, 0.3, 0.5, 0.75, 1, 1.5, 2, 3, 5, 7.5, 10])
```
kasperrisager, line 103: Since this returns a `NelsonSiegel`, you might as well call it `NelsonSiegel`.

src/NelsonSiegelSvensson.jl
```diff
@@ -0,0 +1,179 @@
...
+    # τₐₗₗ = [0.1, 0.15, 0.2, 0.3, 0.5, 0.75, 1, 1.5, 2, 3, 5, 7.5, 10] move to parameter to allow more freedom
+    Δₘ = vcat([maturities[1]], diff(maturities))
+    #sp = sum(prices .* prices)
+    total_resid = Inf
+    ns_param = NelsonSiegel(1.0, 0.0, 0.0, 1)
+
+    for τ in τₐₗₗ
+        spot(m, param) = zero(NelsonSiegel(param[1], param[2], param[3], param[4]), m)
+        param₀ = [1.0, 0.0, 0.0]
+        res = LsqFit.curve_fit(spot, maturities, yields, Δₘ, param₀)
```
kasperrisager, line 113: Again, I don't have experience with `LsqFit`, but I think it does some advanced stuff. In this case, however, you're doing a linear least squares, so maybe there's a simpler way. I'm not sure, but give it a thought.

kasperrisager:
Looks to be going in the right direction 😄. I added some specific comments inside the code. I think at this point, it would be good to add the Nelson Siegel (Svensson) curves into the actual package by including the file in the `Yields.jl` file, and exporting the methods that should be exported. Then, see if you can get the tests to run, and add some tests for the curves. In the first instance, just make some test examples that demonstrate that the code actually runs, and then you can improve them later.

leeyuntien:
Changed more and some tests added. Please review thanks.

src/NelsonSiegelSvensson.jl
```diff
@@ -0,0 +1,181 @@
...
+```
+"""
+struct NelsonSiegelSvensson <: ParametricModel
+    β₀
+    β₁
+    β₂
+    β₃
+    τ₁
+    τ₂
+    #r² no need
```
alecloudenback, line 50: Can the commented lines with this (misc lines) be removed then? How come it's not needed?

src/NelsonSiegelSvensson.jl
```diff
@@ -0,0 +1,181 @@
...
+    #function NelsonSiegelSvensson(β₀, β₁, β₂, β₃, τ₁, τ₂, r²)
+    function NelsonSiegelSvensson(β₀, β₁, β₂, β₃, τ₁, τ₂)
+        #(τ₁ <= 0 || τ₂ <= 0 || r² < 0 || r² > 1) && throw(DomainError("Wrong parameter ranges"))
+        (τ₁ <= 0 || τ₂ <= 0) && throw(DomainError("Wrong parameter ranges"))
+        #return new(β₀, β₁, β₂, β₃, τ₁, τ₂, r²)
+        return new(β₀, β₁, β₂, β₃, τ₁, τ₂)
+    end
+end
+
+Base.zero(ns::NelsonSiegel, t) = Continuous(ns.β₀ .+ ns.β₁ .* (1.0 .- exp.(-t ./ ns.τ₁)) ./ (t ./ ns.τ₁) .+ ns.β₂ .* ((1.0 .- exp.(-t ./ ns.τ₁)) ./ (t ./ ns.τ₁) .- exp.(-t ./ ns.τ₁)))
```
alecloudenback, line 61: The [docs](https://juliaactuary.github.io/Yields.jl/stable/api/#Base.zero-Union{Tuple{YC},%20Tuple{YC,%20Any}}%20where%20YC%3C:Yields.AbstractYield) for `zero` would suggest that `zero(ns::NelsonSiegel,t)` should return a `Periodic(1)` compounding frequency.

The options: 
- `convert(Periodic(1),ns_rate)` to make it consistent
  - and define a `zero(ns::NelsonSiegel,t,c::Continuous)` method for users who want the continuous rate
- We could update the docs - in some sense it doesn't really matter what's returned by default because the `Rate` type conveys that information

I'm open to either option - would returning a `Periodic(1)` here be surprising to users?

src/NelsonSiegelSvensson.jl
alecloudenback: Agree with the thoughts above, but not a big deal to add a somewhat light dependency like this. It looks like everywhere `LsqFit` is used it's qualified with `LsqFit.functionname`. That means instead of `using LsqFit` it could be `import LsqFit` which is probably better code hygiene we should try to use since it doesn't pollute the namespace and it's really clear which functions are defined in other packages.

src/SmoothingSpline.jl
```diff
@@ -0,0 +1,84 @@
...
+    β
+    xs
+
+    function SmoothingSpline(df, β, xs)
+        (df <= 0) && throw(DomainError("Wrong parameter ranges"))
+        return new(df, β, xs)
+    end
+end
+
+using LinearAlgebra
```
alecloudenback, line 21: try to put all the imports at the top of the Yields.jl file

test/NelsonSiegelSvensson.jl
```diff
@@ -0,0 +1,27 @@
+@testset "NelsonSiegelSvensson" begin
+    
+    @testset "NelsonSiegel" begin
+
+        # EURAAA_20191111
```
alecloudenback, line 5: just to confirm - are these test targets from an external validation source (e.g. a textbook, whitepaper, etc.)? It's not clear to me if there is independence of the validating numbers or if that's just the answer from running the code.

src/NelsonSiegelSvensson.jl
```diff
@@ -0,0 +1,181 @@
+abstract type ParametricModel end
```
alecloudenback, line 1: I think this wants to be abstract type ParametricModel <: AbtstractYield end` so that various functions in the rest of the package (e.g. `discount`) will work with these new types via the [fallback methods](https://github.com/JuliaActuary/Yields.jl/blob/master/src/generics.jl).

test/NelsonSiegelSvensson.jl
```diff
@@ -0,0 +1,27 @@
+@testset "NelsonSiegelSvensson" begin
```
alecloudenback, line 1: Please also add some tests of the `zero` function, and ensure that some of the other primary functions like `discount`, `forward`, and `accumulation` work. I think they should since you've defined `zero` and the [fallback methods](https://github.com/JuliaActuary/Yields.jl/blob/master/src/generics.jl) but would be good to have them in the test suite. It would be best if you could validated from an external source what the values of those functions should be, but I think at least testing that they don't error (e.g. `@test forward(ns,1,2) isa Rate`)

alecloudenback (changes requested):
@leeyuntien thank you so much for the contribution! This looks very solid and almost there. Please see @kasperrisager and I's comments. 

If you want a contributor sticker (I still need to send these out to everyone!), send me your address (my username at gmail)

src/NelsonSiegelSvensson.jl
alecloudenback: @kasperrisager any thoughts on this? Abide by returning `Periodic(1)` or just let the rates speak for themselves?

src/NelsonSiegelSvensson.jl
leeyuntien: They are commented out from one of earlier comments from @kasperrisager . They are kept I think until the code is finalized.

src/NelsonSiegelSvensson.jl
leeyuntien: Thank you both. Please advise which way(s) to go, and sure I would love to get a contributor sticker at leeyuntien@gmail.com.

src/NelsonSiegelSvensson.jl
leeyuntien: Agree.

test/NelsonSiegelSvensson.jl
leeyuntien: They are just results from running the code but looking at the curve generated from the parameters they look very close to the actual data. Will try to add more data to test.

leeyuntien: Added more tests.

test/NelsonSiegelSvensson.jl
leeyuntien: Will do.

kasperrisager:
(Just jumping into the discussion since I have the time tonight). Now would also be a good time to `include` your files in /src/Yields.jl and /test/runtests.jl in the same way that is done for the other files. That will allow the tests to be run automatically when you push changes. That will reveal if an error has sneaked in somewhere, and whether the tests cover the new code.

leeyuntien:
Can you advise how to include files into Yields and runtests? You mean push to the main branch, or? Thanks.

alecloudenback:
> Can you advise how to include files into Yields and runtests? You mean push to the main branch, or? Thanks.

He means that currently the package doesn't actually know to grab the code you wrote. The main package needs to [reference the code ](https://github.com/JuliaActuary/Yields.jl/blob/5d1d404338af409a2dd54c1669c01b49623c6d20/src/Yields.jl#L14-L20) like it does for the other parts of the package. Without `include(...)`ing the file, the code doesn't get loaded when the package does.

And the [tests are also organized](https://github.com/JuliaActuary/Yields.jl/blob/5d1d404338af409a2dd54c1669c01b49623c6d20/test/runtests.jl#L4-L9) in a similar fashion.

leeyuntien:
> > Can you advise how to include files into Yields and runtests? You mean push to the main branch, or? Thanks.
> 
> He means that currently the package doesn't actually know to grab the code you wrote. The main package needs to [reference the code ](https://github.com/JuliaActuary/Yields.jl/blob/5d1d404338af409a2dd54c1669c01b49623c6d20/src/Yields.jl#L14-L20) like it does for the other parts of the package. Without `include(...)`ing the file, the code doesn't get loaded when the package does.
> 
> And the [tests are also organized](https://github.com/JuliaActuary/Yields.jl/blob/5d1d404338af409a2dd54c1669c01b49623c6d20/test/runtests.jl#L4-L9) in a similar fashion.

Essentially does this mean I have to modify and push an update of Yield.jl and runtests.jl to include all required files?

kasperrisager:
> Essentially does this mean I have to modify and push an update of Yield.jl and runtests.jl to include all required files?

Yes, exactly. Before you push, you can run the whole test suite locally ([some hints here](https://pkgdocs.julialang.org/v1/creating-packages/index.html#Adding-tests-to-the-package)). That will check exsisting functionality and the new functionality that you have made. Ideally, all the tests will pass before you push, of course, but it's not a strict requirement until the pull request is merged.

leeyuntien:
Thanks. I assume this does not mean to push into the main package but this fork which I have done. I actually included NelsonSiegelSvensson models into Yields and runtests. Please review and advise what else should be done at this stage.

src/NelsonSiegelSvensson.jl
alecloudenback: As we are closer to merging, go ahead and delete them in a new commit - we can always look at prior commits for old code.

alecloudenback:
I pushed a couple of commits to clean up a couple of things. A couple of comments on the changes:

- The tests weren't running properly still and I made edits to allow for this
  - Add new `LsqFit` dependency to `Project.toml`
  - Some of the `@test`s were using the assignment operator (`=`) instead of comparison (e.g. `==` or `≈`)
  - Qualify with the package name `Yields.Nelson...` since the names aren't exported

 Before merging we need more tests for correctness using an external reference.  I added an example of testing the zero rates based on the data here: 

image

However, the tests don't pass as the rates don't match. I'm not sure if I'm grabbing the wrong data or the model needs tweaking still.

alecloudenback:
I pushed the [changes](https://github.com/JuliaActuary/Yields.jl/pull/97/commits) to the branch you are using for this pull request (. I think you can just `git pull` and you will see the commits appear in your history locally.

alecloudenback:
I pulled the rates from the link you left in the comments of the code: https://www.ecb.europa.eu/stats/financial_markets_and_interest_rates/euro_area_yield_curves/html/index.en.html

See the screenshot in [this comment](https://github.com/JuliaActuary/Yields.jl/pull/97#issuecomment-1141634796) for what they looked like on the website:

leeyuntien:
> I pulled the rates from the link you left in the comments of the code: https://www.ecb.europa.eu/stats/financial_markets_and_interest_rates/euro_area_yield_curves/html/index.en.html
> 
> See the screenshot in [this comment](https://github.com/JuliaActuary/Yields.jl/pull/97#issuecomment-1141634796) for what they looked like on the website:

I made slight changes to the tests and found that the results could only generate close zero and par rates but not as close as the tolerances suggest. Are these tolerances industry standards? Can we somewhat loose a bit?

alecloudenback:
I agree with could potentially relax the tolerances (but on the order of basis points, not 10s of basis points). However is it possible that there's something still off with the mechanics? For example, in [this test case](https://github.com/leeyuntien/Yields.jl/blob/efe684f01942d0cd8cb909d767500938669a17ed/test/NelsonSiegelSvensson.jl#L44), the fitted parameters for `β₃`, `τ₁`, and  `τ₂` are very different from what the ECB site indicates (per the screenshot). Is there further tweaking of the fitting code needed?

```
euraaa_zeros = [-0.602009,-0.612954,-0.621543,-0.627864,-0.632655,-0.610565,-0.569424,-0.516078,-0.455969,-0.39315,-0.33047,-0.269814,-0.21234,-0.158674,-0.109075,-0.063552,-0.021963,0.015929,0.050407,0.081771,0.110319,0.136335,0.160083,0.181804,0.201715,0.220009,0.23686,0.252419,0.26682,0.280182,0.292608,0.304191,0.31501]
euraaa_pars = [-0.601861,-0.612808,-0.621403,-0.627731,-0.63251,-0.610271,-0.568825,-0.515067,-0.454526,-0.391338,-0.328412,-0.26767,-0.210277,-0.156851,-0.107632,-0.062605,-0.021601,0.015642,0.049427,0.080073,0.107892,0.133178,0.156206,0.17722,0.196444,0.214073,0.230281,0.245221,0.259027,0.271818,0.283696,0.294753,0.305069]
euraaa_maturities = vcat([0.25, 0.5, 0.75], 1:30)
nss_param = Yields.NelsonSiegelSvensson(0.6348554895516274, -1.2260341163477764, -2.1112390509907732, 0.2927644328902819, 3.0, 1.5)
nss = Yields.est_nss_params(euraaa_zeros, euraaa_maturities)
```

image

I also [added a test set](https://github.com/JuliaActuary/Yields.jl/pull/97/commits/d44dd5f42f6378a58d043939a82b57cba54ee8d6) where it's not focused on fitting the rates, just calculating the output rates. This matches the `zero` rates but the `par` rates error. I will try to look into this.

alecloudenback:

alecloudenback:
I think I botched the rebase - let's continue on #113

leeyuntien:
> This is getting very close! Most of the tests are passing now. I think that the `NelsonSiegel` tests aren't passing because the ECB rates are developed based on `NelsonSiegelSvensson` fit - is that your understanding @leeyuntien ?

Yes that is my understanding so you have to make sure which model you are using in order to match the parameters.

> 
> I pushed some edits. A key thing with the tests/usage is that Yields.jl wants rates as _rates_, not percentages. So for example: for a 1.25% yield, the values should be given as `0.0125` whereas the values that were in the tests would use a value like `1.25`).
> 
> Here's what I think still needs to be done before merging:
> 
> * [ ]  If my assertion above is correct: Figure out an alternative test set for the basic `NelsonSielgel` (ie something other than the ECB rates since it's not testing the same curve)
> * [ ]  Add more details to the docstring.
>   
>   * [x]  What are the tau parameters?
>   * [x]  Do the beta parameters have a meaning
>   * [ ]  Add constructor examples using rates/maturities
> * [x]  Resolve the case where the `discount(curve,time)` errors when `time = 0`  (I added a test for this)
> * [ ]  Delete the parameter comparisons if it's not testing an externally validating test or a property of the curve
> * [ ]  @alecloudenback to rebase onto master before merging and change `<: AbstractYield` to `<: AbstractYieldCurve`
> * [ ]  Include SmoothingSplines?

I saw in src probably we should remove SmoothingSplines and keep SmoothingSpline. You mean to create a separate pull request for this, right? Any others with higher priority to do before doing it?

>   
>   * I suggest splitting those out into a separate PR so that it can be tidied up and tested without blocking the merging of the NelsonSiegel curves

PR #113, opened 2022-06-12
Add NelsonSeigel Curves

alecloudenback:
This is a rebased version of #97 per @leeyuntien, incorporating the recent changes to `master`

Here's what I think still needs to be done before merging:

- [ ] If my assertion above is correct: Figure out an alternative test set for the basic `NelsonSielgel` (ie something other than the ECB rates since it's not testing the same curve)
- [ ] Why do the [parameters](https://github.com/JuliaActuary/Yields.jl/blob/ec24f7092602beebcb70de1d880fed7f41c2c603/test/NelsonSiegelSvensson.jl#L30-L36) differ from the parameters derived in the other package? 
- [ ] Add more details to the docstring.
  - [x] What are the tau parameters?
  - [x] Do the beta parameters have a meaning
  - [ ] Add constructor examples using rates/maturities 
- [x] Resolve the case where the `discount(curve,time)` errors when `time = 0`  (I added a test for this) 
- [x] Delete the parameter comparisons if it's not testing an externally validating test or a property of the curve
- [x] @alecloudenback to rebase onto master before merging and change `<: AbstractYield` to `<: AbstractYieldCurve`
- [ ] Include SmoothingSplines?
  - I suggest splitting those out into a separate PR so that it can be tidied up and tested without blocking the merging of the NelsonSiegel curves 

@leeyuntien - I also modified the fitting constructor to optimize the `τ`s instead of picking from one of the `τ`'s in the list. Please let me know if I should revert that change.

Post-merging: 
- [ ] Reduce dependencies (currently have Roots, Optim, and LsqFit).
- [ ] NelsonSeigel takes a set of zero rates which is similar to the `Yields.Zero` interface. Should the API be `Yields.Zero(Bootstrap(),rates,times)` and `Yields.Zero(NelsonSeigel(),rates,times)`?


alecloudenback:
> NelsonSeigel takes a set of zero rates which is similar to the Yields.Zero interface. Should the API be Yields.Zero(Bootstrap(),rates,times) and Yields.Zero(NelsonSeigel(),rates,times)?

I think we want the API to be something like this. And we'll be able to support more forms of constructor yields. I'll work on this now.

alecloudenback:
closed because I messed up the branch name and I can't push without losing changes (see #116 for replacement)

Issue #114, opened 2022-06-13
Should Yields.Par have a method without `maturities`?


alecloudenback:
The other constructors have that.

Issue #70, opened 2021-12-03
Add Nelson-Siegel and Smoothing Spline Yield Curve fitting


alecloudenback:
Seems like a good reference: 
- https://www.bis.org/publ/bppdf/bispap25.pdf
- https://onriskandreturn.com/2019/12/01/nelson-siegel-yield-curve-model/

Three related variants:
- Nelson-Siegel
- Nelson-Seigel-Svensson
- Smoothing Splines

This could be good first issue for someone familiar with the above methods already.

alecloudenback:
Related article: https://www.soa.org/sections/financial-reporting/financial-reporting-newsletter/2022/february/fr-2022-02-perelman/

leeyuntien:
Would like to contribute to this issue. Please advise next steps. Thanks!

kasperrisager:
A good first step would be to make a struct that holds the data for the Nelson-Siegel parametrisation and methods for computing discount and zero rate correctly. Add in some tests. You can get inspired by the Smith-Wilson implementation. The open a draft pr.

leeyuntien:
Tried to implement and added NelsonSiegelSvensson.jl and SmoothingSplines.jl whose approach might be debatable but please advise how these can be uploaded/pulled for a review.

alecloudenback:
Awesome! Check out this article about creating a pull request: https://www.digitalocean.com/community/tutorials/how-to-create-a-pull-request-on-github

A PR will allow the automated tests/checks to occur and makes it easy to review the changes.

alecloudenback:
Two of three will be completed in #116

leeyuntien:
I have seen there are two spline methods in bootstrapping, do we still need smoothing splines? In this reference https://www.bis.org/publ/bppdf/bispap25.pdf it seems to talk about cubic splines, so smoothing splines here would mean cubic splines, or splines with even higher degrees? Thanks.

> Seems like a good reference:
> 
> * https://www.bis.org/publ/bppdf/bispap25.pdf
> * https://onriskandreturn.com/2019/12/01/nelson-siegel-yield-curve-model/
> 
> Three related variants:
> 
> * [x]  Nelson-Siegel
> * [x]  Nelson-Seigel-Svensson
> * [ ]  Smoothing Splines
> 
> This could be good first issue for someone familiar with the above methods already.

alecloudenback:
[...]

PR #116, opened 2022-06-13
Add NelsonSeigel Curves (per @leeyuntien) and some accomodating refactoring

### LINKED ISSUE #118: Add conversion methods `Continuous(r<:Periodic)` and `Periodic(c<:Continuous,period)`
A convenient alternative to `convert`

alecloudenback:
This is a rebased version of (#97,#113) per @leeyuntien, incorporating the recent changes to `master`

Stems from #70. Closes #118, #114

Here's what I think still needs to be done before merging:

@leeyuntien - can you start with these?

- [x] If my assertion above is correct: Figure out an alternative test set for the basic `NelsonSielgel` (ie something other than the ECB rates since it's not testing the same curve)
- [x] Why do the [parameters](https://github.com/JuliaActuary/Yields.jl/blob/ec24f7092602beebcb70de1d880fed7f41c2c603/test/NelsonSiegelSvensson.jl#L30-L36) differ from the parameters derived in the other package? 
- [x] Include SmoothingSplines?
  - I suggest splitting those out into a separate PR so that it can be tidied up and tested without blocking the merging of the NelsonSiegel curves 

I'll try to work on these:

- [x] Add more details to the docstring.
  - [x] What are the tau parameters?
  - [x] Do the beta parameters have a meaning
- [x] Resolve the case where the `discount(curve,time)` errors when `time = 0`  (I added a test for this) 
- [x] Delete the parameter comparisons if it's not testing an externally validating test or a property of the curve
- [x] @alecloudenback to rebase onto master before merging and change `<: AbstractYield` to `<: AbstractYieldCurve`
- [x] Refactor other methods to allow NS/NSS for different constructor types
  - [x] Add documentation for choice of methods to all the methods now defined at the bottom of `generics.jl`
  - [x] Complete initial sweep of refactor and get tests to where they were earlier
  - [x] Update README docs
  - [x] Why are there cases where `Par` with NSS errors without being able to converge?

@leeyuntien - I also modified the fitting constructor to optimize the `τ`s instead of picking from one of the `τ`'s in the list. Please let me know if I should revert that change.

Post-merging: 
- [ ] Reduce dependencies (currently have Roots, Optim, and LsqFit).


leeyuntien:
> This is a rebased version of (#97,#113) per @leeyuntien, incorporating the recent changes to `master`

I see you have rebased here.

> 
> This is getting very close! Most of the tests are passing now. I think that the `NelsonSiegel` tests aren't passing because the ECB rates are developed based on `NelsonSiegelSvensson` fit - is that your understanding @leeyuntien ?

Yes that is my understanding so you have to make sure which model you are using in order to match the parameters.

> 
> I pushed some edits. A key thing with the tests/usage is that Yields.jl wants rates as _rates_, not percentages. So for example: for a 1.25% yield, the values should be given as `0.0125` whereas the values that were in the tests would use a value like `1.25`).
> 
> Here's what I think still needs to be done before merging:
> 
> @leeyuntien - can you start with these?
> 
> * [ ]  If my assertion above is correct: Figure out an alternative test set for the basic `NelsonSielgel` (ie something other than the ECB rates since it's not testing the same curve)

Can we not use the same curve to test, resulting in different parameter sets?

> * [ ]  Why do the [parameters](https://github.com/JuliaActuary/Yields.jl/blob/ec24f7092602beebcb70de1d880fed7f41c2c603/test/NelsonSiegelSvensson.jl#L30-L36) differ from the parameters derived in the other package?

I think it may be due to different algorithms used in LsqFit and the public package referenced here.

> * [ ]  Include SmoothingSplines?

I saw in src probably we should remove SmoothingSplines and keep SmoothingSpline. You mean to create a separate pull request for this, right?

>   
>   * I suggest splitting those out into a separate PR so that it can be tidied up and tested without blocking the merging of the NelsonSiegel curves
> 
> I'll try to work on these:
> 
> * [x]  Add more details to the docstring.
>   
>   * [x]  What are the tau parameters?
>   * [x]  Do the beta parameters have a meaning
> * [x]  Resolve the case where the `discount(curve,time)` errors when `time = 0`  (I added a test for this)
> * [x]  Delete the parameter comparisons if it's not testing an externally validating test or a property of the curve
> * [x]  @alecloudenback to rebase onto master before merging and change `<: AbstractYield` to `<: AbstractYieldCurve`
> * [ ]  Refactor other methods to allow NS/NSS for different constructor types
>   
>   * [ ]  Add documentation for choice of methods to all the methods now defined at the bottom of `generics.jl`
>   * [ ]  Update README docs
>   * [ ]  Why are there cases where `Par` with NSS errors without being able to converge?
> 
> @leeyuntien - I also modified the fitting constructor to optimize the `τ`s instead of picking from one of the `τ`'s in the list. Please let me know if I should revert that change.
> 
> Post-merging:
> 
> * [ ]  Reduce dependencies (currently have Roots, Optim, and LsqFit).

alecloudenback:
@leeyuntien I saw you just commented on the other PR, let's use this one because I messed up the prior two. 

I'm very excited about this PR and we are close :). The good contribution has made me rethink how some of the other parts of the package works and this is making the code better, more flexible, and generic.

Of the items in the PR, I think getting independent, passing cases in place for `NelsonSiegel` would be a good thing for you to work on @leeyuntien. Since the ECB curves make good test cases for `NelsonSiegelSvensson`, I don't think we need to do much here (though the `Par` function is errorring and I will try to figure that out).

From the perspective of what you wrote previously, a couple of key changes:
- There's now a parameter type used to define the fit parameters (different than the post-fitting curve parameters) 
  - This type is also used to dispatch `Zero`, `Par`, etc just like the bootstrapped set of curves
- I've split out the curve fitting into separate functions and made it more generic to allow us to fit different types of given rates (previously, it assumed and could only fit if given `zero` spot rates).
- Instead of picking from the list of taus, the tau is also solved for with non-linear optimization.

alecloudenback:
We crossed comments there :)

> Can we not use the same curve to test, resulting in different parameter sets?

It is the same curve, but the fit is a bit different so only some of the timepoints match:
image

alecloudenback:
I've debugged a few cases of where `par` was erroring - it seems like the issue is that the newton rootfinding method used isn't robust enough to converge on a rate. `ActuaryUtilities.jl` has a more robust routine for this, but it's a dependency of `Yields.jl`. I need to think about how best to push that functionality up the dependency chain (e.g. into some sort of "Core" package.

alecloudenback:
I've been trying to establish the correctness of the `NelsonSiegel` test cases... and having a very difficult time:

- The **YieldCurve** R package doesn't work on recent R and hasn't been updated in nearly 10 years
- The `nelson_siegel_svensson` package seems to 
  - solve for somewhat close parameters on the doc example, but not exact
  - fail on the ECB test case:
```python
from nelson_siegel_svensson import NelsonSiegelSvenssonCurve
import numpy as np
from nelson_siegel_svensson.calibrate import calibrate_ns_ols

euraaa_maturities = np.concatenate((np.array([0.25, 0.5, 0.75]), np.arange(1, 31, 1.0)))
euraaa_zeros = np.array([-0.602009,-0.612954,-0.621543,-0.627864,-0.632655,-0.610565,-0.569424,-0.516078,-0.455969,-0.39315,-0.33047,-0.269814,-0.21234,-0.158674,-0.109075,-0.063552,-0.021963,0.015929,0.050407,0.081771,0.110319,0.136335,0.160083,0.181804,0.201715,0.220009,0.23686,0.252419,0.26682,0.280182,0.292608,0.304191,0.31501]) / 100

c,s = calibrate_ns_ols(euraaa_zeros,euraaa_maturities)
```

returns parameters that are obviously not correct: 

```
NelsonSiegelCurve(beta0=-29690434.43469232, beta1=29690441.25844614, beta2=29733246.909257106, tau=2.0)
```

- The python code [here](https://onriskandreturn.com/2019/11/13/nelson-siegel-model-python-source-code/) uses a fixed set of `tau`s to pull from and so our method of optimizing tau comes up with different answers.


I've widened the test cases tolerance for these two cases:
- comparing the `NelsonSiegel` fit against the ECB data, which uses NSS to 10bps
- comparing the `NelsonSiegel` fit against the nelson_siegel_svensson packages's demo case to 30bps

Ideally we would have tighter test cases but I haven't found comparables I can make sense of.

alecloudenback:
I've also removed the SmoothingSplines from this PR. @leeyuntien I would love to see another one, and that PR should be much simpler, hopefully! This one took a while before I could focus on it, and then:

1. getting the tests up and running
2. figuring out what the right API was
3. refactoring some of the existing code to accommodate a better API and updating associated documentation
4. investigating these test issues

Tomorrow I'll review the code coverage, one more look at the changes in aggregate, and then merge!

leeyuntien:
Apologies there are so many threads floating around. Can you advise what are your advices on issues still outstanding for me to do on NelsonSiegelSvensson, since you have merged it into master? If not I am starting to work on SmoothingSpline.

> I've also removed the SmoothingSplines from this PR. @leeyuntien I would love to see another one, and that PR should be much simpler, hopefully! This one took a while before I could focus on it, and then:
> 
> 1. getting the tests up and running
> 2. figuring out what the right API was
> 3. refactoring some of the existing code to accommodate a better API and updating associated documentation
> 4. investigating these test issues
> 
> Tomorrow I'll review the code coverage, one more look at the changes in aggregate, and then merge!

alecloudenback:
All good! NS and NSS are done, thanks!

PR #131, opened 2022-07-04
Let curve inherit type of given rates for its methods

### LINKED ISSUE #107: Make yields inherit type if given something other than `Float64`
For example, if user provides input rates that are `Float32` then the yield curve could inherit that type. I think right now it'd get promoted to `Float64` in all cases

alecloudenback:
closes #107

### DIFF
--- a/src/Rate.jl
+++ b/src/Rate.jl
@@ -38,7 +38,9 @@ end
 """ 
     Periodic(frequency)
 
-A type representing periodic interest compounding with the given frequency
+A type representing periodic interest compounding with the given frequency. 
+
+`frequency` will be converted to an `Integer`, and will round up to 8 decimal places (otherwise will throw an `InexactError`). 
 
 # Examples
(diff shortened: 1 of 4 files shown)


alecloudenback:
Per M Abbot on slack: 

> A common pattern here is to make a 2nd type parameter, so that you can get the eltype in function signatures easily:
> 
> ```julia
> struct MyVec{T,A}
>    values::A
>    MyVec(x::A) where {A<:AbstractVector{T}} where T = new{T,A}(x)
> end
> MyVec([1,2,3])
> fun(x::MyVec{T}) where T = T
> ```

instead of 

```julia
struct MyVec{T}
   values::AbstractVector{T}
end
```

alecloudenback:
Closing: this cannot be rebased onto the current code. Numeric-type inheritance is tracked in #107, and #297 covers part of it.

PR #146, opened 2023-02-11
Refactor to `curve` constructor 

### LINKED ISSUE #145: API Update
Sketching out revisions to API that would be breaking:

### Status Quo

- Currently, there are constructors that dictate how the given `Float64` arguments should be interpreted.
  -  E.g. `Yields.Par` vs Yields.Forward`, etc.:

```julia
riskfree_maturities = [0.5, 1.0, 1.5, 2.0]
riskfree    = [5.0, 5.8, 6.4, 6.8] ./ 100     #spot rates, annual effective if unspecified

rf_curve = Yields.Zero(riskfree,riskfree_maturities)
```

  - `SmithWilson` is the only curve type that uses the `ObservableQuotes`:
``` julia
	SmithWilson(zcq::Vector{ZeroCouponQuote}; ufr, α)
    SmithWilson(swq::Vector{SwapQuote}; ufr, α)
    SmithWilson(bbq::Vector{BulletBondQuote}; ufr, α) 
```

### Idea

New API would be:

```julia
curve(method::CurveMethod,quotes::Observable)::T<:AbstractRateCurve
```

so the example above would instead be something like:

```julia
riskfree_maturities = [0.5, 1.0, 1.5, 2.0]
riskfree    = [5.0, 5.8, 6.4, 6.8] ./ 100     #spot rates, annual effective if unspecified
zr = ZeroRate.(riskfree,riskfree_maturities)

rf_curve = curve(Bootstrap(),zr)
```

To some extent, we should be able to automatically convert data to minimize the methods that need to be specified, e.g. a `ZeroRate`(Quote) implies a `ZeroCouponBond`(Quote) price
### Benefits

I think there are some benefits to this:

- Converge API for `SmithWilson` and all of the other constructor types
- More uniform API able to handle different types of "quotes", whether that's yields or prices.
- More extensible for users to define custom curve constructors. Just need to extend the `curve` method for their new `CurveMethod` which can dispatch based on the `quotes` type instead of hooking into a bunch of methods (`Zero`, `Par`, etc...)
- I've been playing with a `Cashflow` type in ActuaryUtilities which would be very similar in concept to the current `ZeroCouponQuote` where the amount of interest is stored with the corresponding timepoint. So if `Cashflow` pans out (I'm having a hard time handling scalars, generators, and vectors), there would be a very similar feel between packages.

cc @kasperrisager

alecloudenback:
closes #145

Status: 

Todo

- [x] Instruments
- [x] add Bond iterator and associated `_pv` 
- [x] Refactor `Bootstrap`
- [ ] Refactor `SmithWilson`
- [ ] Refactor NS/NSS
- [x] Document how to migrate code
	- [ ] removed compounding frequency argument in `zero` (just wrap result that gets returned instead)
- [ ] Document that to automatically assign ordered frequencies to use the non-broadcasted version of the instrument
- [ ] Scrub and update documentation
- [ ] Tests
	- [ ] 

Some misc thoughts:

- How to best represent composite cashflows (e.g. Bond cashflows)? As a `Composite` or a `Vector` of cashflows? 
  - The former seems easier to develop for - being able to condense each problem down to handling cashflows
  - Is the latter more convenient overall? 
  - If `Composite`, then turning a vector into composite is just `reduce(+,v)`. How to conveniently do the opposite? 
  - Some testing notes: `Composite` is faster and `isbits` for small sets, but blows up for even fairly small (n=10) composite. Best to represent as a vector, I think
 
Maybe:
- [ ] Change `FinanceCore.CompoundingFrequency` to just `Frequency` since it's used by `Instruments` for payment frequency?
- [ ] Upstream the instruments/observables into separate package or FinanceCore
- [ ] Bring back `Step` curve? or does `Forward` suffice?
- [ ] Use Weak Dependencies to, e.g. only pull in interpolation methods for `Bootstrap`
- [ ] Add `UnitCashflow` as a specialized `Cashflow`?

To get input on:
- [ ] should the name be `Instruments` or `Observables` for the provided financial instruments?


alecloudenback:
Superseded by #150

PR #152, opened 2023-07-10
DateModel

alecloudenback:
Contracts already use FinanceCore.Timepoint... which is simply `Union{Date,<:Real)`. This PR on the main PR would:

- Define a `DateModel` which has two fields: a `model` (same model you would otherwise use in a `Projection` and `valuation_date`
- in various logic, use dispatch combining the underlying type within the contract and the `model` to determine the decimal time (yearfrac) relative to the val date in the `DateModel`
- Use BusinessDays, DayCounts, and logic from Miletus to determine dates of cashflows for real date bonds


alecloudenback:
Closing: this predates the package restructure. Date support is tracked in #155 under the newer Timeline design.

PR #198, opened 2025-08-03
add details about the interpretation of adding/substracting curves

alecloudenback:


alecloudenback:
test failure relates to a Makie extension, added tests pass

PR #201, opened 2025-08-04
use Newton for bootstrapping

alecloudenback:
This isn't working because the derivatives aren't right - I think it needs a custom `__loss_single_function` or just otherwise custom loss function to allow autodiff.


alecloudenback:
closed in favor of #203

PR #258, opened 2026-05-21
simulate: promote path eltype across model parameters for AD

alecloudenback:
## Summary

`simulate` previously typed each scenario's `cumulative` rate array by `typeof(_sim_initial_rate(model))` only. For Vasicek and CIR that's the type stored in `m.initial`; for Hull-White it comes from the calibrating curve (typically `Float64`). When *another* scalar parameter became a `ForwardDiff.Dual` — e.g. taking `ForwardDiff.derivative(σ -> simulate(HullWhite(a, σ, curve)))` — the step function returned a `Dual` whose write into the `Float64` array crashed at `setindex!`.

This blocked AD-through-Monte-Carlo for σ (vega), mean reversion `a`, and CIR's long-term mean `b` — even though `simulate` is documented to thread `Dual` types.

## Fix

Add `_sim_param_eltype(m)` per model that promotes the scalar field types:

```julia
_sim_param_eltype(m::ShortRate.Vasicek) =
    promote_type(typeof(m.a), typeof(m.b), typeof(m.σ))
_sim_param_eltype(m::ShortRate.CoxIngersollRoss) =
    promote_type(typeof(m.a), typeof(m.b), typeof(m.σ))
_sim_param_eltype(m::ShortRate.HullWhite) =
    promote_type(typeof(m.a), typeof(m.σ))
```

Use it to widen the path-array eltype:

```julia
r0 = _sim_initial_rate(model)
T = promote_type(typeof(r0), _sim_param_eltype(model))
```

Behavior is unchanged for `Float64` callers — the `promote_type` is a no-op when every parameter is concrete `Float64`. The fix only changes the previously-crashing case.

## Test plan

New `@testset "AD through simulate (parameter sensitivities)"` in `test/Stochastic.jl` exercises:

- Hull-White σ (the original failure)
- Vasicek σ
- Vasicek `a` (mean reversion)
- CIR σ

Each asserts `isfinite(ForwardDiff.derivative(…))`. Without the fix, the Hull-White σ test raises `MethodError: convert(::Type{Float64}, ::ForwardDiff.Dual{…})`. With the fix, all four pass. Full Stochastic test set (430 tests) passes.

Adds `ForwardDiff` to `test/Project.toml` (test-only dep; no change to the package's own dep tree).


alecloudenback:
Closing as superseded: #297 promoted `simulate` path element types across model parameters and added ForwardDiff-through-simulation tests.

PR #263, opened 2026-05-26
Collapse ZeroRateCurve into a factory function

alecloudenback:
## Summary

`ZeroRateCurve` is no longer a concrete struct. It is now a factory function that constructs the appropriate underlying curve type via `Yield.build_model`:

- Default spline (`MonotoneConvex`) → returns `Yield.MonotoneConvex`
- Other splines (Linear, PCHIP, Cubic, …) → returns `Yield.Spline`

User-facing API unchanged: `ZeroRateCurve(rates, tenors)`, `ZeroRateCurve(rates, tenors, Spline.Linear())`, and `ZeroRateCurve(curve::AbstractYieldModel, tenors)` all continue to work and produce a curve supporting the full `AbstractYieldModel` interface (`discount`, `zero`, `forward`, callable `(t)`, `pv`, …).

## Why

`ZeroRateCurve` was originally a struct with a lazy per-call rebuild in `discount`:

```julia
function FinanceCore.discount(zrc::ZeroRateCurve, t)
    if t <= zero(t); return one(eltype(zrc.rates)); end
    curve = Yield.build_model(zrc.spline, zrc.tenors, zrc.rates)
    return discount(curve, t)
end
```

That existed to support an `ActuaryUtilities` AD pathway that tagged `zrc.rates` with `ForwardDiff` duals — the lazy rebuild propagated duals through every `discount` call. AU has since moved off that path (`Yield.TenorShift` bumps in AU v5.7+, see [ActuaryUtilities #139](https://github.com/JuliaActuary/ActuaryUtilities.jl/pull/139)), so the rebuild is dead overhead.

After this PR:
- The per-call rebuild is gone — `discount(zrc, t)` is a single spline evaluation.
- The hierarchy has one fewer concrete `AbstractYieldModel` subtype.
- The direct-construction path now mirrors the `fit(...)` path, which already returns `Yield.Spline` / `Yield.MonotoneConvex` directly.

## Breaking changes

Code that dispatched on `::ZeroRateCurve` or accessed `zrc.rates` / `zrc.tenors` / `zrc.spline` will break.

Survey of impact:
- **AU v5.6 internals** dispatched on `::ZRC`. AU v5.7 (in [#139](https://github.com/JuliaActuary/ActuaryUtilities.jl/pull/139)) removed those as part of unrelated unification work, so AU master is fine.
- **AU 5.6.4 (registered)** still has the old dispatches and will not load against FM 5.6.0. AU 5.7 must release before this lands.
- **AU `test/runtests.jl`** had one `zrc.tenors` reference; the paired AU PR updates it to use a local `tenors` variable.
- **No other ecosystem consumers** were found that dispatch on or field-access ZRC.

## Other changes

- Removed `ActuaryUtilities` from `test/Project.toml`. The only file using it (`test/ActuaryUtilities.jl`) is already commented out in `runtests.jl`, and keeping the dep was forcing the registered (incompatible) AU 5.6.4 to precompile at test-setup time.
- Updated `docs/src/index.md` and `docs/src/interpolation.md` to describe ZRC as a factory function.

## Test plan

- [x] All FM tests pass locally — including the 38-test `ZeroRateCurve` testset, which exercises the constructor and the full `AbstractYieldModel` interface (no migration needed; the tests never dispatched on the struct or accessed fields).
- [x] All AU tests pass locally against this FM (after the paired AU PR's one-line `zrc.tenors` migration).
- [ ] CI green.

Version bumped to **5.6.0** — minor (breaking for type-dispatch / field-access; non-breaking for abstract interface consumers).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FinanceModels"
 uuid = "77f2ae65-bdde-421f-ae9d-22f1af19dd76"
-version = "5.5.1"
+version = "5.6.0"
 authors = ["Alec Loudenback <alecloudenback@gmail.com> and contributors"]
 
 [deps]
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -113,7 +113,7 @@ The models can be used to compute various rates of interest:
 
 #### `ZeroRateCurve` — direct construction
 
-Construct a zero-rate curve directly from rates and tenors, without fitting:
+Construct a yield curve directly from zero rates and tenors, without fitting:
 
 ```julia
 rates = [0.02, 0.025, 0.03, 0.035, 0.04]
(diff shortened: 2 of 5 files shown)


alecloudenback:
superseded by #267

PR #270, opened 2026-06-18
MonotoneConvex: fix knot 0/0 NaN + g0==0 discontinuity, unbreak the public fit paths

alecloudenback:
Makes `Yield.MonotoneConvex` reliably fittable to par quotes through the public API, and fixes the two degenerate-forward boundary cases in the Hagan–West construction.

## 1. Knot `0/0` singularity in `g`/`g_rate` — `g1 == 0` (the core bug)
When two adjacent discrete forwards coincide (any flat-forward segment), the interpolated node forward equals the discrete forward, so the sector-(iii) deviation `g1` is exactly `0` (η = 0). The zero-rate integral `G` is evaluated at `x = 0` **exactly at every interior knot**, where sector (iii) divides by η → an unguarded `0/0`.

Two failure layers:
- **value**: `discount(curve, knot) = NaN`;
- **gradient**: still `NaN` after guarding the value, because `iszero(η)` is `false` under ForwardDiff (η carries nonzero partials despite a zero value), so the singular branch runs anyway. This NaN gradient is what silently stalled `fit` at its initial guess for *every* optimizer — the "fitted" rates came back byte-identical to the seed.

Fix: guard on `iszero(x)`. `x` is the normalized interval position — exactly `0` at knots and never the AD variable — so the test is AD-safe. `g(0)=g0` and `G(0)=0` are exact identities (and the correct η→0⁺ limits), so non-degenerate inputs are unchanged.

> Supersedes the "seed from the quotes" theory: the ramp seed merely *triggers* the singularity (it makes `fᵈ[3]==fᵈ[4]`). Reseeding masks it for one dataset but leaves the bug for any flat-forward input or user-built curve, and doesn't stop the optimizer stalling.

## 2. Mirror case — `g0 == 0` (zero-rate discontinuity)
The symmetric degeneracy: when the *left* node forward equals the discrete forward (`g0 == 0`), the interval routed to sector (iii) with η = 3 > 1, where `∫₀¹ g ≠ 0`, so it failed to reprice its right knot — a small (~1e-3) zero-rate **discontinuity** (a grid sweep confirms `g0 == 0` is the *only* region where `∫₀¹ g ≠ 0`). Fixed by classifying `g0 == 0` as sector (ii) (η = 1, `G ≡ 0` on `[0,1)`) via inclusive bounds, plus an `x == 1` boundary guard mirroring `x == 0` (sector (ii) is singular at η = 1, and `x == 1` *is* evaluated at the last knot since `i_time` is capped). `g(1)=g1`, `G(1)=0` are exact identities for every valid sector.

## 3. `Spline.MonotoneConvex` fit paths
`fit(Spline.MonotoneConvex(), q)` and `… Fit.Bootstrap()` errored with `MethodError: Yield.Spline(::Spline.MonotoneConvex, …)`. Both now route to the dedicated native fit (its documented "dispatches to `Yield.MonotoneConvex`" behavior).

## 4. Accessors / ConstructionBase compatibility
- `__default_optic(::MonotoneConvex)` **threw** `iterate(::ClosedInterval)` on first use (broadcast over a non-iterable interval) → now one bounded optic per knot.
- `MonotoneConvex` wasn't reconstructable (cached `f`/`fᵈ`, only a 2-arg constructor) so `@set`/`setproperties` failed → added a `constructorof` that rebuilds from `rates`/`times`, recomputing the cache.

## Verification
- Short-tenor par fit now matches a square-system Newton reference to ~2e-13; **0 NaN/Inf gradients across 200 random rate vectors** (incl. forced equal-adjacent cases).
- `g0 == 0` grid-sweep repricing violations: **24 → 0**; the adversarial flat-segment curve's knot discontinuity drops from ~9e-4 to ~3e-10.
- New regression tests close the AD-gradient coverage gap that let this ship (the suite tested values, never gradients).
- Full `Pkg.test` passes — Monotone Convex 496/496, spline/bootstrap/ZeroRateCurve/audit/Aqua all green, **zero regressions**.

---
⚠️ **Base note:** this branch is stacked on the unmerged v6.1 work (`perf/zrc-eager-build`), so the diff also includes those 9 commits.


alecloudenback:
Superseded by #271, which is branched off released `master` (v6.1.0) and contains **only** the MonotoneConvex fix (3 files, +139/−4). The v6.1 commits bundled into this PR already shipped via #266/#267, so this PR's base is now stale. Closing in favor of the clean standalone PR.

PR #273, opened 2026-06-19
Silence spurious SecondOrder ADtype warning; consolidate fit-loss AD backend (v6.2.0)

alecloudenback:
## Summary

Two related changes to yield-curve **fitting** (`src/fit.jl`), released as **v6.2.0** (minor):

1. **Fix** — silence a spurious `OptimizationBase` warning emitted on *every* `fit` call when a second-order optimizer is used.
2. **Refactor** — collapse the duplicated fit-loss plumbing behind one AD backend and one shared loss builder.

Both are numerically inert: default-optimizer fits are **byte-identical** before and after.

## 1. The fix

Fitting with a second-order optimizer printed this on every call:

```
┌ Warning: The selected optimization algorithm requires second order derivatives, but
│ `SecondOrder` ADtype was not provided. So a `SecondOrder` with AutoForwardDiff() for
│ both inner and outer will be created ...
└ @ OptimizationBase .../cache.jl:49
```

**Root cause:** the loss builders declared the AD as first-order `AutoForwardDiff()`. With a second-order optimizer (`Newton()` on the unbounded MonotoneConvex path; `IPNewton()` on the bounded generic path), OptimizationBase needs a Hessian, finds only a first-order declaration, auto-promotes to `SecondOrder(AutoForwardDiff(), AutoForwardDiff())`, and warns once per fit. A market-data build (~18 curves) prints it ~18×.

**Fix:** declare `SecondOrder` AD explicitly (it's exactly what auto-promotion was already building). The default first-order optimizers — `LBFGS` and `Fminbox(LBFGS)` — only request **gradients**, computed via `SecondOrder`'s *inner* `AutoForwardDiff`, so the Hessian capability is declared but never instantiated for them.

## 2. The refactor

The fix would have left `DifferentiationInterface.SecondOrder(AutoForwardDiff(), AutoForwardDiff())` repeated at three call sites, and two of the loss builders were near-identical. Consolidated to:

- **`const __FIT_ADTYPE`** — one source of truth for the fitting AD backend.
- **`__reprice_loss(build, loss_method, quotes)`** — shared builder; reprices `quotes` under `build(u)` (parameters → model), building the model **once per loss evaluation** (unchanged from before). `__monotone_convex_loss_function` and `__spline_loss_function` are now one-line delegations; the generic `fit` closure keeps its AccessibleModels transform but uses `__FIT_ADTYPE`.

Net **−11 lines** in the loss plumbing; also removes a `p`-shadowing footgun in the old spline loss.

## Why it's safe (verified)

Controlled A/B on the default path (same loss, only the declared AD differs), solved with `LBFGS`:

| Path | Δ solution | Δ allocations |
|---|---|---|
| MonotoneConvex | `0.0` | `0 B` |
| generic `Yield.Constant` | `0.0` | `0 B` |
| generic `Yield.NelsonSiegel` | `0.0` | `0 B` |

Post-refactor fits are **byte-identical to the pre-change residuals** and warning-free:

| Fit | resid | warnings |
|---|---|---|
| `MonotoneConvex` + `Newton` | `3.697e-14` | 0 |
| `MonotoneConvex` + `LBFGS` (default) | `1.442e-10` | 0 |
| `NelsonSiegel` + `IPNewton` | `1.634e-3` | 0 |
| `CairnsPritchard` + `IPNewton` | `6.807e-11` | 0 |

The forward-over-forward **Hessian** consumed by second-order optimizers is finite and symmetric even at the `#271` degenerate-forward singularities (`g0/g1 == 0` flat-forward configs) — checked directly, not just via repricing.

Full local suite across every fit path (`Yield` incl. spline fit, `Equity`, `CairnsPritchard`, `NelsonSiegelSvensson`, `MonotoneConvex`, `Stochastic`, `regressions`) — all green, unchanged pass counts.

## Note on the bounded generic path

Plain `Newton()` is unusable on the bounded generic path regardless of this change — Optim's `Fminbox` rejects it (`ArgumentError: Newton is not supported as the Fminbox optimizer`), thrown during setup before the AD-cache warning. `IPNewton()` is the bounds-compatible second-order method, and it is the one that surfaced (and is fixed by) this change.

## Tests

- `test/MonotoneConvex.jl` — `Newton()` fits (both `Yield.` and `Spline.MonotoneConvex` entry points) are quiet (`@test_logs`) and accurate.
- `test/regressions.jl` — generic `IPNewton()` path and the `MonotoneConvex` `Newton()` path.

The "no warning" assertion uses the codebase's dependency-free `@test_logs` idiom (default `min_level = Warn`); confirmed it genuinely catches a warning.

## Release

`Project.toml` → **6.2.0** (minor; skips the 6.1.2 patch — `v6.1.1` is already tagged). No `[compat]` changes (`DifferentiationInterface`/`AutoForwardDiff` are already direct deps).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FinanceModels"
 uuid = "77f2ae65-bdde-421f-ae9d-22f1af19dd76"
-version = "6.1.1"
+version = "6.2.0"
 authors = ["Alec Loudenback <alecloudenback@gmail.com> and contributors"]
 
 [deps]
--- a/src/fit.jl
+++ b/src/fit.jl
@@ -294,7 +294,14 @@ function fit(
     end
     amodel = AccessibleModel(Base.Fix2(neg_loss_fn, quotes), mod0, variables)
     tf = AccessibleModels.transformed_func(amodel)
-    optf = Optimization.OptimizationFunction((x, p) -> convert(eltype(x), -tf(x, p)), AutoForwardDiff())
+    # Declare second-order AD so a bounds-compatible second-order optimizer (e.g.
+    # `IPNewton()`) finds the Hessian it needs instead of triggering OptimizationBase's
+    # auto-promotion warning. The default `LBFGS()` only requests gradients (via the
+    # inner `AutoForwardDiff`), so the Hessian capability is unused — no change for it.
+    optf = Optimization.OptimizationFunction(
+        (x, p) -> convert(eltype(x), -tf(x, p)),
+        DifferentiationInterface.SecondOrder(AutoForwardDiff(), AutoForwardDiff())
+    )
     x0 = collect(AccessibleModels.transformed_vec(amodel))
     bounds = AccessibleModels.transformed_bounds(amodel)
     lb = haskey(bounds, :lb) ? collect(bounds.lb) : nothing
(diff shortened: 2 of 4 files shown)


alecloudenback:
Bumped `Project.toml` to **6.1.2** (patch) — `v6.1.1` is already tagged, so this fix ships as the next patch release. No `[compat]` changes (uses `DifferentiationInterface.SecondOrder` + `AutoForwardDiff`, both already direct deps).

alecloudenback:
Updated: added the fit-loss refactor (single `__FIT_ADTYPE` backend + shared `__reprice_loss` builder) and changed the bump to **6.2.0** (minor, skipping 6.1.2) — supersedes the 6.1.2 note above. Refactor is byte-identical on the default path; full suite green. History force-pushed to drop the interim 6.1.2 commit.

PR #280, opened 2026-07-15
Add FX module: covered-interest-parity forward curves, contracts, and quote conventions (v6.3.0)

alecloudenback:
## Summary

Adds foreign-exchange support as a new `FX` module. The design observation: under covered interest parity, an FX forward curve is a ratio of discount factors scaled by spot —

$$F(t) = S_0 \cdot \frac{DF_f(t)}{DF_d(t)}$$

— so FX needs **no new curve mathematics**. Every existing spline, parametric model, bootstrap, curve-arithmetic operator, and `fit` path applies to FX curve construction unchanged. What this PR adds is the semantic layer where FX errors actually happen: explicit pair direction, loud failures on mismatched or inverted currencies, and market-quote constructors that refuse to guess conventions.

## What's included

- **`FX.Pair(:EUR, :USD)`** — a currency pair as a plain value struct (base/quote, matching market "EURUSD" naming); `inv` flips it. Currencies can be any values — `Symbol`s, strings (plain, `SubString`, or the InlineStrings codes CSV readers produce), or ISO 4217 numeric codes — and pairs compare by **content**: `==`/`hash` delegate to the fields, so string storage widths never split a code, while `Symbol` vs. string stays distinct. Pricing any contract against a model for a different (or inverted) pair throws an `ArgumentError` naming both pairs. Degenerate pairs (`FX.Pair(:USD, :USD)`) are deliberately allowed: the identity pair lets generic multi-currency code route domestic contracts through the same conversion machinery.
- **`FX.Forwards(pair, spot, domestic, foreign)`** — the CIP model: `forward(m, t) = spot * discount(foreign, t) / discount(domestic, t)`, callable as `m(t)`; `inv(m)` is the flipped-pair model and is calibration-exact (`forward(inv(m), t) == 1/forward(m, t)` identically, no refit). Construction requires a strictly positive, finite spot. There is deliberately no two-time `forward(m, from, to)`: under CIP the fair forward for delivery at `to` is independent of contracting time.
- **`FX.Forward(pair, strike, time)`** — outright forward contract; `pv = (F(t) − K)·DF_d(t)`, with a `cur_time` form (only the discounting date moves — the model is not rolled; settled forwards are worth zero, matching the projection filter convention).
- **`FX.Outright(pair, F, t)`** — market forwards as zero-price `Quote`s (entering at market costs nothing — the FX analog of par-bond quoting). Forward *points* deliberately have **no constructor**: a guessed pip scale is the classic silent JPY error, and a required one only renames the arithmetic, so points convert explicitly at the data boundary — `FX.Outright.(pair, spot .+ points ./ 10_000, ts)` (`1/100` for JPY-quoted pairs).
- **`FX.implied_zcb_quotes`** — the load-bearing transform: given spot and the domestic curve, each forward quote pins the implied base-currency discount factor in closed form, `DF_f(t) = (K·DF_d(t) + price)/S₀`, feeding existing curve construction (same architecture as QuantLib's `FxSwapRateHelper`, in ~10 lines). Mixed-pair sets and quotes implying non-positive/non-finite discount factors are refused with `ArgumentError`s naming the offending quote.
- **`FX.ParBasisSwap(pair, spread, maturity; reference)`** — par cross-currency basis-swap spread quotes for calibrating the long-dated basis. Under standard collateralized quoting the domestic leg pays its own discount curve flat (worth par — it drops out), so each quote materializes into a deterministic base-currency strip `Quote(1.0, FX.BasisSwapLeg(...))` against the `reference` projection curve, replicating `Bond.Floating`'s fix-in-advance projection exactly. These flow through the same `fit` methods as outrights, so **one calibration mixes forward outrights (short end) and basis swaps (long end)**. Non-whole tenors form a short first stub (the schedule stays anchored at maturity, per `Bond.coupon_times`) that accrues its *actual* length, compound-accrued in `ParYield`'s stub convention — a zero-spread leg on its own curve stays exactly par; irregular periods anywhere else are out of scope.
- **`fit` integration** — one-step `fit(FX.Forwards(pair, spot, dom, Spline.Linear()), quotes, Fit.Bootstrap())` (and `Fit.Loss`) for spline foreign curves via the per-quote implied transform; parametric foreign curves (e.g. `Yield.Constant`, NS/NSS) calibrate through the generic optimizer path via optics composed through the `foreign` field.
- **`FX.Converted(contract, pair, key)`** — cross-currency projection wrapper: converts each projected cashflow into the quote currency at the CIP forward for its time, with the FX model resolved from the projection's model store under `key` (the `Bond.Floating` key/value pattern — a converted floating leg resolves its own reference curve from the same store, so cross-currency swaps are ordinary `Composite`s). The declared `pair` is checked against the model found under the key, so an **inverted model under an otherwise-valid key errors** instead of silently converting at reciprocal rates; a non-FX model under the key is likewise named at the source.

**Present values follow one rule**: pv is denominated in the currency of the contract's own cashflows — quote currency for `FX.Forward` (it settles in quote units), base currency for `FX.BasisSwapLeg` (a base-currency strip). This preserves the package-wide repricing invariant `pv(fitted, q.instrument) ≈ q.price`; convert with spot before combining across denominations.

## Cross-currency basis / hedge pricing

The `foreign` curve is *defined by what reprices the FX market*, not as any particular reference curve, which handles the post-2008 CIP basis structurally:

- **Absorbed**: fitting to market forwards makes the fitted `foreign` curve the basis-adjusted (CSA/collateralized) discount curve — hedge pricing with the fitted model is basis-consistent by construction.
- **Explicit**: `FX.Forwards(pair, spot, usd_ois, eur_ois + basis)` via existing zero-rate curve arithmetic, and conversely `basis = m.foreign - eur_ois` extracts it as a first-class curve.

## Textbook and independent validation

- Hull (*Options, Futures, and Other Derivatives*), two editions: §5.10 CIP (AUD/USD forward 0.6453) and the later-edition Example 5.6 restatement (0.7206, both covered-interest-arbitrage profits); the §7.9 fixed-for-fixed currency swap ($1.5430M, Table 7.9 forwards, both valuation routes agreeing) and the later Examples 7.2/7.3 ($0.9629M, with the printed B_D/B_F bond values).
- Basis-swap calibration validated four independent ways: a hand-derived flat case (flat 3% − 20bp ⇒ CSA exactly flat 2.8%), an analytic par-spread round-trip on a non-flat truth curve, mixed short/long-end bootstrap repricing, and the actual composite swap (converted floating EUR leg vs. spot-scaled floating USD leg) pricing to zero through the independent projection/store machinery.

## Out of scope (documented follow-ups)

1. The mark-to-market (resetting-notional) basis-swap *contract*, pending convention decisions (settlement timing, accrual on the adjustment). Valued at CIP forwards under deterministic curves, the resetting (flat domestic) leg is a telescoping strip of one-period par loans — each at the domestic curve's own forward, each worth zero — so the constant-notional and MTM par spreads are **exactly** equal (asserted to machine precision in the test suite) and quoted MTM spreads calibrate the constant-notional representation without approximation. The two structures differ only by a convexity term on the resetting leg requiring *both* an index–discount spread on that leg *and* stochastically correlated FX and rates — the standard treatment (Fujii, Shimada & Takahashi 2010, §3.5) values the MtM leg as a portfolio of one-period swaps that is worth exactly zero when its index is the discounting rate.
2. Multi-pair consistency: independently fitted pair models are not forced to satisfy triangle arbitrage. The scaling design — documented on the FX page — is a per-currency ("collateral-consistent") market object: one spot and one curve per currency, every pair's forwards derived, `N` curves instead of up to `N(N−1)/2` pair models, with `FX.Forwards` as the pairwise view it returns.
3. Garman–Kohlhagen option methods (structurally `Equity.BlackScholesMerton(r_d, r_f, σ)`); delta-conventioned volatility surfaces; stochastic FX for `simulate`.

## Testing

- `test/FX.jl` (134 assertions): CIP forwards and inversion (including pricing under `inv(m)`), contract pricing with `cur_time` semantics, content-equality across currency representations (Symbol / String / SubString / integer ISO codes), quote conventions and broadcasting, closed-form implied-DF round-trips against a non-flat bootstrapped domestic curve, off-market quote handling, bootstrap and `Fit.Loss` spline fits, parametric fits via composed optics, explicit-basis composition, `Converted` conversion/identity/cur_time/Eduction cases, both Hull parameterizations, all four basis-swap validation routes, front-stub tenor correctness (single-stub and stub-schedule cashflows, telescoping par, and calibration through a stub-tenor quote), mark-to-market equivalence (each one-period loan of the resetting leg par to machine precision; the MTM domestic leg nets the inception exchange; the full MTM swap prices to zero at the constant-notional par spread — whole-period and stub schedules), rolled-valuation-date leg pricing, and the guard suite (pair mismatches, inverted model under a valid key, mis-keyed store, non-positive spot, `df ≤ 0` quotes, same-maturity clashes).
- Full suite green locally (Julia 1.12.5), including Aqua. Strict docs build clean; new guide page `docs/src/fx.md` + `API/FX.md`.

## Notes for review

- **Version**: bumped 6.2.0 → 6.3.0. This trivially collides with #278 (also claiming 6.3.0) on the Project.toml version line — whichever merges second takes the usual mechanical rebase.
- `FX.Forward` (contract) vs. the top-level `Forward` time-shift wrapper: name collision avoided by module qualification, consistent with `Bond.Fixed`/`Option.EuroCall` convention. `FX.Pair` shadows `Base.Pair` only *inside* the FX module (deliberate; `=>` is unaffected), following the market term "currency pair".
- No precompile-workload additions, keeping load-time impact nil.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FinanceModels"
 uuid = "77f2ae65-bdde-421f-ae9d-22f1af19dd76"
-version = "6.2.0"
+version = "6.3.0"
 authors = ["Alec Loudenback <alecloudenback@gmail.com> and contributors"]
 
 [deps]
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ Commonly, we deal with conventions that imply a contract and an observed price. 
 - `ParYield`
 - `ParSwapYield`
 - `ForwardYield`
-- `FX.Outright` and `FX.ForwardPoints` (foreign-exchange forwards)
+- `FX.Outright` (foreign-exchange forwards)
 - `FX.ParBasisSwap` (par cross-currency basis-swap spreads, for calibrating the long-dated basis)
 
 FinanceModels offers a way to define new contracts as well.
--- a/README.md
+++ b/README.md
@@ -67,6 +67,7 @@ Existing `struct`s:
 - `Option.EuroCall` and `Option.EuroPut`
 - `CommonEquity`
 - `Option.Cap`, `Option.Floor`, `Option.Swaption` (interest rate derivatives)
+- `FX.Forward` (an outright foreign-exchange forward)
 
 Commonly, we deal with conventions that imply a contract and an observed price. For example, we may talk about a treasury yield of `0.03`. This is a description that implies a `Quote`ed price for an underling fixed bond. In FinanceModels, we could use `CMTYield(rate,tenor)` which would create a `Quote(price,Bond.Fixed(...))`. In this way, we can conveniently create a number of `Quote`s which can be used to fit models. Such convenience methods include:
 
@@ -76,6 +77,7 @@ Commonly, we deal with conventions that imply a contract and an observed price. 
 - `ParYield`
 - `ParSwapYield`
 - `ForwardYield`
+- `FX.Outright` and `FX.ForwardPoints` (foreign-exchange forwards)
 
 FinanceModels offers a way to define new contracts as well.
 
@@ -104,6 +106,10 @@ Stochastic models include:
 
 - `ShortRate.Vasicek`, `ShortRate.CoxIngersollRoss`, and `ShortRate.HullWhite` short-rate models
 - `Equity.BlackScholesMerton` for option valuation
+
+Foreign exchange models include:
+
+- `FX.Forwards` — covered-interest-parity forward exchange rates from a spot rate and a discount curve per currency, fit to forward/points quotes (see the Foreign Exchange docs page)
 
 #### Yield-related functions
 
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ Existing `struct`s:
 - `CommonEquity`
 - `Option.Cap`, `Option.Floor`, `Option.Swaption` (interest rate derivatives)
 - `FX.Forward` (an outright foreign-exchange forward)
+- `FX.Converted` (a contract's cashflows converted into another currency at forward FX rates, e.g. for cross-currency swaps)
 
 Commonly, we deal with conventions that imply a contract and an observed price. For example, we may talk about a treasury yield of `0.03`. This is a description that implies a `Quote`ed price for an underling fixed bond. In FinanceModels, we could use `CMTYield(rate,tenor)` which would create a `Quote(price,Bond.Fixed(...))`. In this way, we can conveniently create a number of `Quote`s which can be used to fit models. Such convenience methods include:
 
--- a/README.md
+++ b/README.md
@@ -79,6 +79,7 @@ Commonly, we deal with conventions that imply a contract and an observed price. 
 - `ParSwapYield`
 - `ForwardYield`
 - `FX.Outright` and `FX.ForwardPoints` (foreign-exchange forwards)
+- `FX.ParBasisSwap` (par cross-currency basis-swap spreads, for calibrating the long-dated basis)
 
 FinanceModels offers a way to define new contracts as well.
 
@@ -110,7 +111,7 @@ Stochastic models include:
 
 Foreign exchange models include:
 
-- `FX.Forwards` — covered-interest-parity forward exchange rates from a spot rate and a discount curve per currency, fit to forward/points quotes (see the Foreign Exchange docs page)
+- `FX.Forwards` — covered-interest-parity forward exchange rates from a spot rate and a discount curve per currency, fit to forward/points quotes and/or par basis-swap spreads (see the Foreign Exchange docs page)
 
 #### Yield-related functions
 
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -21,6 +21,7 @@ makedocs(;
             "Contracts" => "contracts.md",
             "Rates" => "Rates.md",
             "Yield Curve Arithmetic" => "yield_arithmetic.md",
+            "Foreign Exchange" => "fx.md",
             "Migration Guide" => "migration.md",
         ],
         "Modules" => [
@@ -31,6 +32,7 @@ makedocs(;
             "Yield" => "API/Yield.md",
             "Bond" => "API/Bond.md",
             "Equity" => "API/Equity.md",
+            "FX" => "API/FX.md",
             "Option" => "API/Option.md",
             "Volatility" => "API/Volatility.md",
             "ShortRate" => "API/ShortRate.md",
--- a/docs/src/API/FX.md
+++ b/docs/src/API/FX.md
@@ -1,0 +1,21 @@
+# FinanceModels.FX API Reference
+
+```@index
+Modules = [FinanceModels.FX]
+```
+
+## Exported API
+
+```@autodocs
+Modules = [FinanceModels.FX]
+Private = false
+```
+
+## Unexported API
+
+```@autodocs
+Modules = [FinanceModels.FX]
+Public = false
+```
+
+Please [open an issue](https://github.com/JuliaActuary/FinanceModels.jl/issues) if you encounter any issues or confusion with the package.
--- a/docs/src/fx.md
+++ b/docs/src/fx.md
@@ -47,8 +47,11 @@ returning a wrong number.
 
 Forward *points* are the second convention trap. The interbank market quotes forwards as
 pips over spot, and the pip size differs by pair: `0.0001` for most pairs but `0.01` for
-JPY-quoted pairs. [`FX.ForwardPoints`](@ref) therefore takes the scale explicitly rather
-than guessing (see below).
+JPY-quoted pairs. There is deliberately no points constructor: a function that guessed
+the scale would be the classic silent JPY error, and one that required it would only
+rename the arithmetic. Convert points explicitly at the data boundary, where the
+pair-dependent pip factor stays visible at the call site:
+`FX.Outright.(pair, spot .+ points ./ 10_000, times)`.
 
 ## Covered interest parity
 
@@ -111,8 +114,9 @@ bond quoting at 1.0:
 ```julia
 quotes = FX.Outright.(eurusd, [1.1055, 1.1113, 1.1225, 1.1459], [0.25, 0.5, 1.0, 2.0])
 
-# equivalently, from points over spot (pips; scale is required: 10_000 here, 100 for JPY quotes):
-quotes = FX.ForwardPoints.(eurusd, [55.0, 113.0, 225.0, 459.0], [0.25, 0.5, 1.0, 2.0]; spot = 1.10, scale = 10_000)
+# equivalently, from screen points over spot (pips: 1/10_000 here, 1/100 for JPY quotes):
+points = [55.0, 113.0, 225.0, 459.0]
+quotes = FX.Outright.(eurusd, 1.10 .+ points ./ 10_000, [0.25, 0.5, 1.0, 2.0])
 ```
 
 ## Constructing a curve from market quotes
@@ -288,8 +292,8 @@ estr = ...   # EUR projection curve, fitted from EUR OIS quotes
 sofr = ...   # USD discount curve
 
 quotes = [
-    # short end: forward points
-    FX.ForwardPoints.(eurusd, [25.0, 55.0, 120.0], [0.25, 0.5, 1.0]; spot = 1.10, scale = 10_000);
+    # short end: outrights from screen points over spot (pips = 1/10_000)
+    FX.Outright.(eurusd, 1.10 .+ [25.0, 55.0, 120.0] ./ 10_000, [0.25, 0.5, 1.0]);
     # long end: par basis-swap spreads
     FX.ParBasisSwap.(eurusd, [-0.0012, -0.0015, -0.0018], [2.0, 5.0, 10.0]; reference = estr)
 ]
--- a/docs/src/fx.md
+++ b/docs/src/fx.md
@@ -313,6 +313,21 @@ Deliberately *not* built yet: the mark-to-market (resetting-notional) basis-swap
 *contract* itself, pending convention decisions (which leg resets, settlement timing,
 accrual on the adjustment). Open an issue if your use case needs it sooner.
 
+Also deliberately deferred: **multi-pair consistency**. Each `FX.Forwards` is a
+self-contained model of one pair, and independently fitted pair models are not forced
+to be mutually consistent — nothing makes a directly fitted EURJPY model agree with the
+EURJPY forwards implied by triangulating EURUSD × USDJPY, and each fit implies its own
+version of any shared currency's discount curve. The scaling design is a per-currency
+(*collateral-consistent*) market object: pick one numéraire/collateral currency, store
+one spot rate and one collateral-consistent discount curve per currency, and *derive*
+every pair's forwards from those. For ``N`` currencies that is ``N`` curves instead of
+up to ``N(N-1)/2`` independent pair models, and triangle consistency holds by
+construction rather than by calibration. The tradeoff is exact repricing: where the
+market quotes a direct cross basis inconsistent with triangulation, such a model
+reprices the cross at its derived arbitrage-consistent forward, not at the screen
+quote. `FX.Forwards` is the two-currency case of that design and would become the
+pairwise view a market object returns, so nothing in the current API forecloses it.
+
 ## Relationship to other models
 
 - **FX options**: the Garman–Kohlhagen formula is `Equity.BlackScholesMerton(r_domestic,
--- a/docs/src/fx.md
+++ b/docs/src/fx.md
@@ -249,10 +249,60 @@ basis_swap = Composite(
 principal exchange at maturity is represented; a par basis swap's initial exchange nets
 against a unit `Cashflow` at time zero if you need it explicitly.)
 
-Deliberately *not* built yet, pending convention decisions: mark-to-market (resetting-
-notional) basis swaps, and par-basis-swap `Quote` constructors for calibrating the
-long-dated basis directly from swap quotes — forward-points quotes cover the liquid short
-end today. Open an issue if your use case needs these sooner.
+### Calibrating the long-dated basis from par basis-swap quotes
+
+Forward points cover the liquid short end (out to a year or two); beyond that the
+cross-currency basis trades as **par basis-swap spreads** — "5y EUR/USD basis −18bp"
+means the five-year swap of €STR − 18bp against SOFR flat, with notional exchanges,
+prices to par. [`FX.ParBasisSwap`](@ref) turns those quotes into `fit` inputs.
+
+The quoting assumption, standard for collateralized (OIS-discounted) swaps, is that the
+domestic leg pays the forwards of the same curve used for domestic discounting, so it is
+worth par and drops out. What remains is a condition on the foreign side alone:
+
+```math
+\sum_i (f_i + b)\,\delta\,DF_f(t_i) + DF_f(T) = 1
+```
+
+where ``f_i`` are forwards from the foreign *projection* curve (e.g. a fitted €STR
+curve — a known input passed as the `reference` keyword), ``b`` is the quoted spread,
+``\delta`` the accrual fraction, and ``DF_f`` the basis-adjusted discount curve being
+calibrated. Because the coupon amounts are fully determined by the projection curve,
+each quote materializes into a deterministic base-currency cashflow strip
+([`FX.BasisSwapLeg`](@ref)) that must price to par on the curve being fit — and
+deterministic strips flow through the same fitting machinery as the implied zero-coupon
+quotes. One `fit` call takes both quote types:
+
+```julia
+estr = ...   # EUR projection curve, fitted from EUR OIS quotes
+sofr = ...   # USD discount curve
+
+quotes = [
+    # short end: forward points
+    FX.ForwardPoints.(eurusd, [25.0, 55.0, 120.0], [0.25, 0.5, 1.0]; spot = 1.10);
+    # long end: par basis-swap spreads
+    FX.ParBasisSwap.(eurusd, [-0.0012, -0.0015, -0.0018], [2.0, 5.0, 10.0]; reference = estr)
+]
+
+m = fit(FX.Forwards(eurusd, 1.10, sofr, Spline.Linear()), quotes, Fit.Bootstrap())
+```
+
+Every quote — outright or swap — reprices under the fitted model, and the composite
+basis swap from the previous section values to zero against it (the test suite asserts
+both). Two practical notes:
+
+- Prefer *local* interpolation (e.g. `Spline.Linear`) when bootstrapping coupon-bearing
+  quotes: a global spline reshapes earlier segments as later knots are added, drifting
+  already-solved swaps off par.
+- Interbank basis swaps are quoted on the mark-to-market (resetting-notional)
+  structure. Under deterministic curves the constant-notional and MTM par spreads agree
+  to first order, so quoted spreads calibrate the constant-notional representation
+  directly; the difference only becomes material alongside FX–rates correlation, i.e.
+  with a stochastic FX model.
+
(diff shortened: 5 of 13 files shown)


alecloudenback:
Pushed bbd51c6 addressing review feedback:

**Coverage** — the −0.05% project drop was the two `__fx_foreign_optic` branches (the documented unbounded-optic forms, which no built-in model exercises) — both now unit-tested, plus new error-path coverage (`FX.Converted` missing store key → `KeyError`). Patch coverage should be 100% now.

**Independent textbook validation** — new tests reproduce published numbers from Hull, *Options, Futures, and Other Derivatives* (9th ed.):
- §5.10 CIP forward: AUD/USD spot 0.6200, 5%/7% continuous → 2y forward **0.6453** (plus covered-interest-arbitrage sign checks for off-market strikes).
- §7.9 fixed-for-fixed currency swap (9% USD / 4% JPY flat, ¥110=\$1, receive 5% on ¥1,200M vs pay 8% on \$10M, 3y): reproduces Table 7.9's forwards (0.009557 / 0.010047 / 0.010562), the two bond values (\$9.6439M, ¥1,230.55M), and the swap value **\$1.5430M** via *both* of Hull's valuation routes — and asserts the portfolio-of-forwards route equals the two-bonds-at-spot route to numerical precision.
- Discrete-compounding IRP form `F = S·(1+r_d)ⁿ/(1+r_f)ⁿ` with `Periodic` rates.

**Cross-currency swaps** — added `FX.Converted(contract, key)`: converts each projected cashflow into the quote currency at the CIP forward for its payment time, with the FX model resolved from the projection model store by key (the `Bond.Floating` pattern). A currency swap is then an ordinary `Composite` — the Hull swap above is the test/doc example — and because the wrapped contract still sees the same store, a converted `Bond.Floating` resolves its own reference curve, so constant-notional float-for-float **basis swaps** compose today. The fundamental identity `pv(dom, Converted(c)) == spot·pv(foreign, c)` is asserted for fixed and floating legs. Deliberately deferred pending convention decisions: MTM (resetting-notional) basis swaps and par-basis-swap `Quote` constructors for long-end basis *calibration* (forward points cover the liquid short end).

**Docs** — `fx.md` substantially expanded: conventions/direction/points-scale table, the CIP no-arbitrage argument spelled out, the closed-form implied-DF derivation, cross-currency basis economics (absorbed vs. explicit, hedged-curve arithmetic `asset - foreign + domestic`), the worked Hull swap, and basis-swap composition.

Local: full suite green (FX 58/58, Aqua 10/10), docs build clean under strict cross-reference checks. (`evaluate` failure is pre-existing #217.)

alecloudenback:
`23f1785` addresses the findings from an actuarial code review of this branch:

- **3-arg `present_value` hole (the one fix-before-merge finding)**: `pv(m, fx_forward, cur_time)` previously bypassed both the analytic FX method and the pair-mismatch `ArgumentError` (both were 2-arg only) and fell into the generic projection path's cryptic iteration error — exactly what the guard was written to prevent. Both methods now take `cur_time = 0.0`; a forward settled before `cur_time` values to zero, consistent with the `cf.time >= cur_time` projection filter everywhere else.
- **`fit` dedupe**: the two spline-tag FX fit methods now delegate to one `__fit_fx_via_implied` helper (signatures stay separate for the documented anti-ambiguity reason).
- **`maturity(::FX.Converted)` limitation documented**: undefined for transducer-wrapped (`Eduction`) contracts — deliberately not "fixed," since a transducer may alter cashflow times, making any general delegation silently wrong.
- **`Base.show` for `FX.Pair`**: prints `FX.Pair(:EUR, :USD)` instead of the singleton-type noise in quote displays.
- **Tests 58 → 64**: valuation-time pricing/boundary/settled-forward cases, 3-arg mismatch error, `repr`, and `inv` under non-flat curves.

Deliberately deferred from the review: forward-forward FX rates (`forward(m, from, to)`), and a friendlier error for store-keyed contracts priced without a store (shared with `Bond.Floating`; ecosystem-wide fix belongs in its own PR).

Local verification: full suite green, FX 64/64, Aqua 10/10 (no ambiguities from the new 3-arg methods), strict docs build clean.

alecloudenback:
`c6dc226` adds **`FX.ParBasisSwap`** — par cross-currency basis-swap `Quote`s for calibrating the long-dated basis directly from swap quotes (previously listed as deliberately deferred).

**Design.** Under standard OIS-discounted quoting, the domestic leg pays its own discount curve's forwards flat, so it is worth par and drops out; what remains is a base-currency par condition `Σ(fᵢ+b)·δ·DF_f(tᵢ) + DF_f(T) = 1` whose coupon amounts are fully determined by the foreign *projection* curve (passed as `reference`). The constructor materializes that leg — with exactly `Bond.Floating`'s fix-in-advance convention, pinned by an invariant test — into `Quote(1.0, FX.BasisSwapLeg(pair, cashflows))`, a deterministic strip the existing bootstrap/loss machinery prices directly. A per-quote transform (`FX.__implied_foreign_quote`) lets one `fit` call take a mixed set: forward points for the liquid short end, basis swaps for the long end. This mirrors how QuantLib structures the same problem (known collateral curve, basis on the non-USD leg), while staying store-free per this package's fit architecture.

**Validation** (no published constant-notional bootstrap table exists to borrow, so four independent routes):
- hand-derived flat-curve case (3% − 20bp ⇒ CSA curve exactly flat 2.8%, `DF = 1.028⁻ᵗ`);
- analytic par-spread round-trip (spreads computed on a truth curve; fit recovers it and every quote reprices to `<1e-9`);
- the *actual* composite swap — `Composite(FX.Converted(Bond.Floating(b,…)), −spot·Bond.Floating(0,…))` — values to ~0 under the fitted model through the entirely separate projection/store machinery;
- parametric foreign curve recovery via the generic optic fit.

**Also in this commit**: the later-edition Hull examples now sit alongside the 9th-edition parameterizations in the testset — Example 5.6 (forward 0.7206; arbitrage profits \$21.87/\$55.79) and Examples 7.2/7.3 (Table forwards 0.009182/0.009275/0.009368; B_D \$10.4191M, B_F ¥1,252.01M; swap ≈ \$0.9629M via both valuation routes) — so the tests are verifiable against current printings as well.

Scope note: the MTM (resetting-notional) basis-swap *contract* remains the one deferred item; under deterministic curves its par spreads agree with constant-notional to first order (documented), so quoted spreads calibrate correctly today.

Local verification: full suite green, FX 95/95, Aqua 10/10, strict docs build clean.

alecloudenback:
`156fbb6` makes non-Symbol currency parameters first-class on `FX.Pair`. The struct was already fully parametric — every method dispatches on `Pair{B,Q}` — so this just removes the two Symbol-specific touchpoints:

- The convenience constructor is now unconstrained: anything Julia admits as a type parameter works — `Symbol`s, `isbits` values (e.g. `InlineStrings` codes: `FX.Pair(String3("EUR"), String3("USD"))`, or ISO 4217 numeric codes: `FX.Pair(978, 840)`), or types. A plain `String` is not `isbits` and errors naturally at construction.
- `show` prints `repr` of each parameter: Symbol pairs display exactly as before (`FX.Pair(:EUR, :USD)`), and inline-string pairs round-trip as valid constructor syntax (`FX.Pair(String3("EUR"), String3("USD"))`).

Docs call out the footguns: distinct parameter values are distinct pairs (a `Symbol` pair never matches an inline-string pair, and `String3("EUR")` never matches `String7("EUR")` — relevant since CSV.jl picks inline-string widths per column), with `Symbol(code)` normalization at the data boundary as the alternative. Tests exercise an integer-code pair end-to-end — construction, `inv`, `show`, CIP pricing, and the cross-convention mismatch `ArgumentError` — and the InlineStrings path was verified directly in a scratch environment. FX 101/101, Aqua 10/10, docs build clean.

alecloudenback:
**`d573646` — `FX.Pair` reworked as a value struct, plus fixes from a second review pass.**

`Pair` now carries its currencies as fields (`base`, `quote_currency`) instead of type parameters. The deciding observation: no FX method ever dispatched on a *specific* pair — every method was uniform `where P`, so the pair never selected code; it is data. The singleton design was paying real costs for that non-benefit: a docs paragraph on type-parameter legality, the `String3`≠`String7`≠`Symbol` identity footgun, type-unstable construction from runtime data, and a duplicate error-fallback `pv` method per contract. Pairs now compare by **content** (`==`/`hash` delegate to the fields, so the storage widths CSV readers assign per column never split a currency code), plain `String`s work, and the diagonal+fallback method pairs collapse to single methods throwing the same `ArgumentError`. Symbol pairs still `===` and print identically.

Second-review findings, dispositioned:

- **Accepted — `ForwardPoints` scale is now required.** The docs already claimed the scale was taken "explicitly rather than guessing"; the `10_000` default contradicted that and was the silent JPY off-by-100× footgun.
- **Accepted — `Converted` guards its store lookup.** Verified failure mode: `forward(yield_curve, t)` returns a `Rate`, which *silently Rate-scales* the amount and only errors at `Cashflow` reconstruction deep in the transducer pipeline. Now an `ArgumentError` naming the key and the offending type.
- **Accepted — implied-DF transform rejects `df ≤ 0`** at the offending quote instead of surfacing as NaNs inside a downstream spline fit.
- **Accepted (docs) —** `cur_time` does not roll the model; `inv` of a fitted model is calibration-exact; `ParBasisSwap` front-stub behavior spelled out.
- **Rejected — making `pv(m, ::BasisSwapLeg)` quote-currency.** The review's premise ("the fit path never needs it") is false: parametric-foreign calibrations price `Quote(1.0, leg)` through the *generic* optimizer via `present_value(m, q.instrument)`, so foreign-units pv preserves the load-bearing repricing invariant; `spot * pv` would calibrate the curve to `1/spot` × par. The denomination rule is documented in the module docstring (0d9ec51).
- **Rejected — two-time `forward(m, from, to)`.** The proposed formula (`S·F(to)/F(from)`) is not a meaningful quantity: under CIP the fair forward for delivery at `to` is `forward(m, to)` regardless of contracting time. It's a clean `MethodError` today (verified: no fallback — `Forwards` sits outside the yield hierarchy); the Forwards docstring now says why the form deliberately doesn't exist.
- **Rejected — rejecting degenerate pairs.** `FX.Pair(:USD, :USD)` with unit spot is the *identity* pair (`forward ≡ 1`), which lets generic multi-currency code route domestic contracts through the same `Converted` machinery — now documented and tested.
- **Rejected — constraining `K` on the `BasisSwapLeg` leaf.** The review's premise is inverted: every leaf `__foldl__` in the package (`Cashflow`, `Bond.Fixed`, `Bond.Floating`) leaves `K` unconstrained; only `asfoldable` wrappers gate on kind. The leg already matches the convention.
- **Test gaps accepted:** pricing under `inv(m)`, `Converted` through the projection path with `cur_time > 0`, plus new guard tests (required scale, mis-keyed store, `df ≤ 0`, content equality across string types, identity pair). FX testset: 116 assertions.

Full suite green, Aqua 10/10, strict docs build clean.

alecloudenback:
**`1fa47a7` — `FX.ForwardPoints` removed** (supersedes the scale-required change in `d573646`).

Rationale: once `scale` was required, the constructor computed nothing but `spot + points / scale`. A guessed scale is the silent JPY off-by-100× error; a required one only renames the arithmetic while the function name claims the convention. The points convention's entire content *is* the pip factor, so it now lives as explicit arithmetic at the data boundary, visible at the call site:

```julia
FX.Outright.(pair, spot .+ points ./ 10_000, times)   # 1/100 for JPY-quoted pairs
```

The pip education moved to the `FX.Outright` docstring and the conventions section of the FX docs page. FX testset 113, strict docs build clean.

alecloudenback:
**`0b48d18` — third-review disposition (all three findings accepted) + PR description rewritten to match the current API.**

- **`Converted` direction blindness — accepted; the strongest finding of the three.** An inverted model under an otherwise-valid key silently multiplied cashflows by reciprocal rates (reviewer's repro: 0.891 instead of 1.122), exactly the bug class the module promises to make loud. `FX.Converted(contract, pair, key)` now carries the declared pair, and projection checks it against the store's model. This partially reverses an earlier claim of mine that the wrapper's blindness had "no local fix": the *model-direction* half is checkable and now checked; the *contract-denomination* half (wrapping a USD bond and calling it EUR) genuinely isn't, since bare cashflows carry no currency — the docstring now frames `pair` as that declaration. Regression test with `inv(fx)` under a valid key included.
- **`ParBasisSwap` front stub — accepted, resolved by rejection.** The reviewer is right that consistency with `Bond.Floating` proved convention-matching, not financial correctness: a 0.1y quarterly "quote" accrued a full 1/4 period. Rather than invent a stub-accrual convention (which day count? which stub end?), non-whole-period maturities now throw — basis-swap quote sheets are whole-tenor instruments, and the package's convention is to refuse to guess. `Bond.Floating`'s own stub behavior is pre-existing and out of scope here.
- **Spot validation — accepted.** An exchange rate is a strictly positive price; there is no legitimate `spot ≤ 0`, and the failure mode is ordinary-looking wrong numbers (negative forwards, sign-flipped pvs), not a crash. Construction now requires positive, finite spot. The reviewer's "preferably finite" caught a real hole: zero spot slipped past the implied-DF `df > 0` guard as `df = Inf` — that guard now requires finiteness too.
- **Stale PR description — fixed.** The body was written at the first commit and still advertised the removed `ForwardPoints`, described `Pair` as a type-level singleton with dispatch-enforced matching, claimed 36 assertions, and listed `Converted`/basis calibration as out of scope. Rewritten to the current surface: value-struct `Pair` with content equality, `Outright`-only quoting with the explicit points idiom, `ParBasisSwap` mixed calibration, `Converted(contract, pair, key)`, the pv-denomination rule, and the guard suite.

FX testset 120; full suite + Aqua green; strict docs build clean.

alecloudenback:
**`1d4fc98` — `ParBasisSwap` front-stub tenors now supported with true accrual** (supersedes the whole-tenor rejection in `0b48d18`, per maintainer direction: accommodate a short first period, no otherwise-irregular schedules).

`Bond.coupon_times` already produces exactly that shape — anchored at maturity, regular `1/f` periods counted backward, front-truncated — so the only irregular period is ever the first. Each coupon now fixes in advance over its **actual** window (`[0, t₁]` for the stub, killing the negative-time lookback) at the simple annualized forward over that window, times the window's true length:

- **Whole periods are unchanged to floating point**: the `f`-periodic rate over a `1/f` window *is* the DF ratio minus one, so the new formula collapses to the old `rate(freq(forward(t−1/f, t)))/f` exactly.
- **The stub compound-accrues**, which is `ParYield`'s documented sub-period convention ("compound-accrued so that `(1 + stub_coupon)·DF = 1`") and buys the telescoping property: a zero-spread leg on its own curve is *exactly* par under any front-stub schedule — asserted at `1e-14`.
- The reviewer's 0.1y example now produces **1.00276** (their expected figure, matching Periodic 3% compound accrual), not the 1.0069 the fabricated full-quarter accrual gave.
- `Bond.Floating`'s full-period stub accrual is pre-existing and left as-is; the strip-equals-projected-floating-leg identity is documented as exact for whole-period tenors. Aligning the core contracts' stub accrual is a candidate follow-up issue, since it changes released behavior.

New tests: single-stub cashflow, 2.5y-annual stub schedule, telescoping par, and calibration through a stub-tenor quote recovering a flat truth CSA curve exactly alongside whole tenors. FX testset 127; full suite + Aqua green; strict docs build clean.

alecloudenback:
## Response to the mark-to-market (P1) review finding, plus a fifth review round

### The P1 claim, adjudicated

The reviewer's structural observation is correct — constant-notional and MTM are different contracts with different cashflows — but the valuation claim ("their par spreads agree only approximately under restricted conditions," with differences driven by "large rate differentials, long maturities, or volatile forward curves") is **incorrect under this package's valuation framework**, and the fix runs in the opposite direction: the docs previously *under*claimed ("agree to first order"). The true statement is **exact equality**, now proved in the docs and asserted in the test suite.

**Why.** In the market-standard MTM convention the *flat domestic* (USD) leg's notional resets to spot each period while the spread leg's notional stays constant. Valued at CIP forwards — which is exact *static replication* under deterministic curves, no expectation or measure argument required — period *i* of the resetting leg is: lend `F(tᵢ₋₁)` at `tᵢ₋₁`, receive it back with interest at the domestic curve's own forward at `tᵢ`. That is a one-period loan at the lender's own curve — worth exactly zero, every period, **regardless of curve shape**. Summing the strip telescopes the notional flows into precisely the MTM leg (inception exchange, interest on the resetting notional, interior reset payments, final notional). So the entire domestic side drops out period by period, under the same quoting assumption the constant-notional transform already makes (domestic leg pays its own discounting curve) — and what remains, the constant-notional foreign leg, is identical in both structures. Same par condition ⇒ same spread ⇒ same calibrated `DF_f`.

**Numerically** (non-flat bootstrapped curves; 5y quarterly and a 5.3y front-stub schedule): max one-period loan PV `2.2e-16`; MTM domestic-leg PV minus spot `0.0` and `4.4e-16`; `b_MTM − b_CN = 0.0`. Committed as the new **"MTM (resetting-notional) equivalence"** testset, so the equality is a regression-tested fact rather than a documentation assertion. Note the claimed drivers — rate differentials, maturity — produce no difference at all: the replication never references curve shape.

**This is also the standard curve-construction treatment.** Fujii, Shimada & Takahashi, *A Note on the Construction of Multiple Swap Curves with and without Collateral* (2010), §3.5, price the MtM-CCS as "a portfolio of the strips of the one-period traditional CCS," confirm the convention (USD flat leg resets; spread leg constant), and show the resetting side's value is **exactly zero when its index is the discounting rate** (their Eqs. 58–59), with the residual otherwise a *covariance* term between FX and the discount bonds arising from the index–discount spread (Eqs. 60–61). That localizes the genuine MTM/CN difference precisely: it requires **both** an index–discount spread on the resetting leg **and** stochastically correlated FX and rates — the first is shut off by the module's stated quoting assumption, the second by deterministic curves. "Volatile forward curves" can only matter through a channel deterministic models do not have, which is exactly why the MTM *contract* remains on the stochastic roadmap.

### On the three proposed remedies

- **Rename to `ConstantNotionalParBasisSwap`** — declined: the quote the constructor accepts *is* the market's par basis-swap spread, and under the documented framework it calibrates the curve without approximation. Encoding an incorrect premise into the API name would be worse than stating the theorem, which the docstring and docs page now do.
- **Require users to pre-convert spreads** — declined: in-model there is nothing to convert, and the convexity residual is not a fixed offset a user could apply without a stochastic model.
- **Implement the MTM contract** — remains deferred, now with a sharper documented reason: under deterministic curves it adds zero calibration information (proved above); it becomes meaningful with stochastic FX/rates, or for projecting the reset settlement flows themselves.

### Also in this round (actuarial code-review findings)

- **`ParBasisSwap` docstring example switched `Spline.Cubic` → `Spline.Linear`** — the docstring demonstrated the interpolation the docs page and tests explicitly warn against for coupon-bearing bootstraps. Measured: 2.7e-6 silent repricing drift on the 2y basis quote under Cubic vs 1e-16 under Linear (single-cashflow outright quotes sit on knots and are exact under any interpolating spline, so the other Cubic examples stand).
- fx.md's par-condition formula now writes `δᵢ` (stub-aware), matching the docstring; the intro's "explicit currency of every cashflow" overclaim reworded (cashflows are currency-blind — the *contract* carries the declaration); new practical note that `Fit.Loss` mixes discount-factor-scale and par-price-scale residuals across quote families (the bootstrap is exact and unaffected).
- New matched-pair `pv(model, leg, cur_time)` assertion — previously only the mismatched-pair guard exercised the three-argument path.
- Follow-up issue #281 opened: aligning `Bond.Fixed`/`Bond.Floating` front-stub accrual with the true-accrual convention (changes released behavior → its own PR + migration note).

Tests: `test/FX.jl` now 134 assertions; full suite + Aqua green locally; strict docs build clean. PR description updated to match.

PR #302, opened 2026-08-24
Add a named zero_rate API

alecloudenback:
Exports zero_rate(curve, time) as the preferred financial operation while retaining Base.zero compatibility. Documents the additive migration path so a future major release can restore the usual Base.zero meaning. Existing third-party implementations continue to work.


alecloudenback:
Closing: `zero(curve, t)` stays the zero-rate API. Its type and arity already separate it from Base's one-argument `zero`, and an alias would give one concept two names. A one-argument additive identity (#82) can still be added later without conflict.

PR #305, opened 2026-09-03
Reject unsuccessful fits and use sloped spline seeds

alecloudenback:
## Summary

- seed loss-fitted spline knot rates with a small slope around the former 5% seed, avoiding PCHIP and Akima divided-difference singularities under ForwardDiff
- require successful Optimization return codes before constructing models in the generic optic, MonotoneConvex, and spline fit paths
- add regression coverage for 10-quote CMT and par fits plus deterministic failed-solver coverage for all three paths
- use the documented nearby seed in the sensitive NSS calibration test now that unsuccessful solves throw

This is stacked on #295 because the spline and MonotoneConvex fit paths use its shared KnotGrid implementation.

## Before and after

Julia 1.12.5, five warmed samples per case; time and allocation are medians. The baseline timings are misleadingly low because Optim stops immediately and fit returns the flat seed.

| Quotes | Spline | Version | Max relative price error | Median time | Median allocation |
|---|---|---:|---:|---:|---:|
| CMT | PCHIP | before | 6.47% | 0.234 ms | 296 KB |
| CMT | PCHIP | after | 4.00e-13% | 0.616 ms | 532 KB |
| CMT | Akima | before | 6.47% | 0.195 ms | 299 KB |
| CMT | Akima | after | 8.66e-13% | 0.526 ms | 551 KB |
| Par | PCHIP | before | 6.47% | 0.217 ms | 296 KB |
| Par | PCHIP | after | 6.44e-13% | 0.622 ms | 532 KB |
| Par | Akima | before | 6.47% | 0.233 ms | 299 KB |
| Par | Akima | after | 1.73e-12% | 0.491 ms | 551 KB |

## Validation

- full Pkg.test suite passed in an isolated environment using a clean FinanceCore checkout
- final targeted NSS and audit regression suites: 437/437 passed
- git diff --check passed

### DIFF
--- a/docs/src/models.md
+++ b/docs/src/models.md
@@ -94,7 +94,7 @@ julia> rate(r)
 
 ### Available Models - Stochastic Short Rates
 
-Stochastic short-rate models with closed-form zero-coupon bond prices. These are full yield models that also support Monte Carlo simulation. See the [Stochastic Models](@ref) guide for details and examples.
+Stochastic short-rate models with closed-form zero-coupon bond prices. These are full yield models that also support Monte Carlo simulation. See the [Stochastic Interest Rate Models](@ref) guide for details and examples.
 
 - [`FinanceModels.ShortRate.Vasicek`](@ref)
 - [`FinanceModels.ShortRate.CoxIngersollRoss`](@ref)
--- a/src/fit.jl
+++ b/src/fit.jl
@@ -255,6 +255,21 @@ function __reprice_loss(build, loss_method, quotes)
     return Optimization.OptimizationFunction(loss, __FIT_ADTYPE)
 end
 
+# Every optimizer-backed fit must either return a successful solution or throw. A
+# failed solve may leave `sol.u` equal to the initial guess, so constructing a model
+# from it would make optimizer failure look like a valid (but unfitted) result.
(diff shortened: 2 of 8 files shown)


alecloudenback:
Follow-up performance check for paths that did not have the PCHIP/Akima silent failure. Each number is the mean of 1,000 warmed public fit calls on Julia 1.12.5.

| Path | Before | After | Change | Allocation change |
|---|---:|---:|---:|---:|
| generic Constant | 27.93 us | 29.14 us | +4.3% (+1.2 us) | +16 B |
| generic ZeroRateCurve | 826.57 us | 816.11 us | -1.3% | +16 B |
| MonotoneConvex | 651.10 us | 659.84 us | +1.3% | +16 B |
| Linear spline | 246.13 us | 252.57 us | +2.6% | +16 B |
| Cubic spline | 426.27 us | 431.15 us | +1.1% | +16 B |

The generic and MonotoneConvex numerical results were identical before and after. Their only code-path change is the successful-retcode check; the MonotoneConvex seed has the same values as before. These timing differences are at or near ordinary run-to-run noise, with the worst relative result occurring on the deliberately tiny 28 us Constant fit. Linear and Cubic use the new sloped seed, but show no material performance regression.

